PR #73, opened 2020-07-12
Bitmap Library

### LINKED ISSUE #53: Paging Improvements
- Separate bitmap library
- Freeing pages
- Stress testing
- Detecting hardware memory limit

### LINKED ISSUE #68: Bitmap Library

Kfeavel:
Resolves #68 
Resolves #53

### DIFF
--- a/kernel/arch/i386/paging_asm.s
+++ b/kernel/arch/i386/paging_asm.s
@@ -1,0 +1,14 @@
+.section .text
+.align 4
+
+.global px_invalidate_page
+px_invalidate_page:
+#if TARGET_MACHINE >= TARGET_MACHINE_I486
+    movl    4(%esp),%eax
+    invlpg  (%eax)
+#else   /* TARGET_MACHINE < TARGET_MACHINE_I486 */
+    movl    %cr3,%eax
+    movl    %eax,%cr3
+#endif
+    ret
+
--- a/kernel/lib/bitmap.cpp
+++ b/kernel/lib/bitmap.cpp
@@ -1,3 +1,23 @@
 #include <lib/bitmap.hpp>
 
+size_t bitmap_find_first_range_clear(bitmap_t *bitmap, size_t size, size_t count) {
+    size_t bm, check_lo, check_hi, check, masked, idx, ofst;
+    size_t mask = ((size_t)1 << count) - (size_t)1;
+    for (size_t i = 0UL; i < size; i++) {
+        idx = INDEX_FROM_BIT(i);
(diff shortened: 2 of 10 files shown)


kernel/arch/i386/paging_asm.s
```diff
@@ -0,0 +1,14 @@
+.section .text
+.align 4
+
+.global px_invalidate_page
+px_invalidate_page:
+#if TARGET_MACHINE >= TARGET_MACHINE_I486
```
Kfeavel, line 6: What is the purpose of this section?

kernel/lib/bitmap.cpp
```diff
@@ -0,0 +1,52 @@
...
+}
+
+size_t bitmap_find_first_bit_clear(bitmap_t *bitmap, size_t size) {
+    for (size_t i = 0; i < size; i++) {
+        if (!((bitmap[INDEX_FROM_BIT(i)] >> OFFSET_FROM_BIT(i)) & 1)) return i;
+    }
+    return SIZE_T_MAX_VALUE;
+}
+
+/*
```
Kfeavel: Do we still need this diagram / should we clean it up to be more like actual documentation?

kernel/main.cpp
```diff
@@ -113,11 +113,34 @@ extern "C" void px_kernel_main(const multiboot_info_t* mb_struct, uint32_t mb_ma
     px_rs232_print("\n");
     px_rs232_print(model);
     px_rs232_print("\n");
-    // Now that we're done make a joyful noise
+
+    px_kprintf(DBG_WARN "Mapping in test pages\n");
```
Kfeavel: Do we want to keep the tests now that we know it works?

kernel/arch/i386/paging_asm.s
micahswitzer: To clear the TLB

kernel/lib/bitmap.cpp
micahswitzer: We should probably get rid of it. We could keep it around for documentation if we want.

kernel/main.cpp
micahswitzer: Maybe we could move it somewhere else

kernel/arch/i386/paging_asm.s
Kfeavel: As per what we discussed, the invalidate page instruction does not exist on anything lower than i486, so this is necessary. Closing.

PR #91, opened 2020-07-31
WIP: UEFI Support

### LINKED ISSUE #90: Add UEFI Support with BootBoot
**Is your feature request related to a problem?**
BIOS is outdated and UEFI will make our lives easier.

**Describe the feature request in detail**
BIOS sucks and UEFI will make our lives much easier for things like getting ACPI tables and native resolutions. Additionally, if we move to x86_64, UEFI will already boot the system into 64 bit mode, so we won't need to do that either.

**Additional context**
BootBoot only supports booting in 64 bit mode, so that might be an issue, or it might not be. Depends on how well Panix runs on x86_64 based systems since it's built for 32 bit architectures.

Kfeavel:
Closes #90 

TODO:
- [ ] Separate makefile architectures into separate files


Kfeavel:
Closed because UEFI requires 64 bit support, which is not available yet. Instead, these changes to the Makefile will be pushed as a separate branch.

PR #99, opened 2020-08-06
ANSI Cursor Position Codes

### LINKED ISSUE #98: Update putchar to handle ANSI cursor position
**Is your feature request related to a problem?**
We should not have a function to print at a given location

**Describe the feature request in detail**
putchar (putc) should handle the ANSI code to move the cursor

**Additional context**
```
Esc[Line;ColumnH
Esc[Line;Columnf
```
**Cursor Position:**
Moves the cursor to the specified position (coordinates).
If you do not specify a position, the cursor moves to the home position at the upper-left corner of the screen (line 0, column 0). This escape sequence works the same way as the following Cursor Position escape sequence.

Kfeavel:
Closes #98 
TODO:

- [x] Save cursor position code
- [x] Restore cursor position code

### DIFF
--- a/kernel/dev/tty/tty.cpp
+++ b/kernel/dev/tty/tty.cpp
@@ -114,7 +114,16 @@ int putchar(char c) {
                 ansi_state = Value;
                 return c;
             }
-            // handle any other control codes here
+            else if (c == 's') { // Save cursor position attribute
+                ansi_cursor_x = tty_coords_x;
+                ansi_cursor_y = tty_coords_y;
+                return c;
+            } 
+            else if (c == 'u') { // Restore cursor position attribute
+                tty_coords_x = ansi_cursor_x;
+                tty_coords_y = ansi_cursor_y;
+                return c;
+            }
             break;
         case Value:
             if (c == ';') { // the semicolon is a value separator
@@ -158,20 +167,6 @@ int putchar(char c) {
                         tty_coords_y = ansi_val;
                     }
                 }
-                ansi_state = Normal;
-                ansi_val = 0;
-            }
-            else if (c == 's') { // Save cursor position attribute
-                ansi_cursor_x = tty_coords_x;
-                ansi_cursor_y = tty_coords_y;
-                // Return to normal
-                ansi_state = Normal;
-                ansi_val = 0;
-            } 
-            else if (c == 'u') { // Restore cursor position attribute
-                tty_coords_x = ansi_cursor_x;
-                tty_coords_y = ansi_cursor_y;
-                // Return to normal
                 ansi_state = Normal;
                 ansi_val = 0;
             }
--- a/kernel/dev/tty/tty.cpp
+++ b/kernel/dev/tty/tty.cpp
@@ -39,6 +39,8 @@ uint16_t ansi_val = 0;
 
 uint8_t tty_coords_x = 0;
 uint8_t tty_coords_y = 0;
+uint8_t ansi_cursor_x = 0;
+uint8_t ansi_cursor_y = 0;
 px_tty_vga_color color_back = VGA_Black;
 px_tty_vga_color color_fore = VGA_White;
 // Colors set by tty_clear()
@@ -120,7 +122,7 @@ int putchar(char c) {
                 PUSH_VAL(ansi_val);
                 ansi_state = Bracket;
                 ansi_val = 0;
-            } else if (c == 'm') { // the color/text attributes command
+            } else if (c == 'm') { // Set color/text attributes command
                 PUSH_VAL(ansi_val);
                 // take action here
                 // iterate through all values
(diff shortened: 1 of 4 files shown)


kernel/dev/tty/tty.cpp
```diff
@@ -96,31 +98,40 @@ static size_t ansi_values_index = 0;
...
 #define POP_VAL() ansi_values[--ansi_values_index]
 
-void putchar(char c) {
+int putchar(char c) {
     switch (ansi_state) {
         case Normal: // print the character out normally unless it's an ESC
             if (c != ESC) break;
             ansi_state = Esc;
-            return;
+            return c;
```
micahswitzer: This may be a good candidate for `goto end` instead of many `return c`s

kernel/dev/tty/tty.cpp
```diff
@@ -142,11 +153,36 @@ void putchar(char c) {
...
                 ansi_val = 0;
+            } else if (c == 'H' || c == 'f') { // Set cursor position attribute
+                PUSH_VAL(ansi_val);
+                // take action here
+                // iterate through all values
+                // the proper order is 'line (y);column (x)'
+                while (ansi_values_index > 0) {
+                    ansi_val = POP_VAL();
+                    // If we only have one item in the stack now (i.e. line is done)
+                    if (ansi_values_index == 1) {
```
micahswitzer: What happens if there were more or less than 2 values provided? Should it always accept them and just only look at the first two pushed?

kernel/dev/tty/tty.cpp
```diff
@@ -188,4 +224,21 @@ void putchar(char c) {
...
         tty_coords_y = X86_TTY_HEIGHT - 1;
     }
+    return c;
+}
+
+int puts(const char *str) {
+    int i = 0;
+    // Loops until a null character
+    while(str[i]) {
+        if(putchar(str[i]) == EOF) {
```
micahswitzer, line 236: Maybe you just haven't finished writing the code yet, but I don't see any code in `putchar` that returns `EOF`.

kernel/main.cpp
```diff
@@ -118,22 +118,24 @@ extern "C" void px_kernel_main(const multiboot_info_t* mb_struct, uint32_t mb_ma
...
     int i = 0;
     while (true) {
         // Display a spinner to know that we're still running.
+        // Each line will save the cursor position, change it
+        // to the top right corner (-1) and then restore the
+        // cursor position (to the saved location).
         switch (i) {
             case 0:
-                px_kprintf("\b|");
+                px_kprintf("\033[s\033[%i;%iH\b|\033[u", X86_IND_Y, X86_IND_X);
```
micahswitzer: I understand wanting to test with this, but maybe it should be merged in separately once this PR is ready?

Makefile
```diff
@@ -218,11 +218,11 @@ vbox: vbox-create
 
 # Open the connection to qemu and load our kernel-object file with symbols
 .PHONY: debug
-debug: dist/panix.iso
+debug: dist/panix.kernel
```
micahswitzer: These changes are unrelated to this PR

kernel/dev/tty/tty.cpp
Kfeavel: > What happens if there were more or less than 2 values provided? Should it always accept them and just only look at the first two pushed?

I think that would be easiest. It would be non-standard to provide more than 2 values, so the safest thing to do is just ignore them. The code will need a little tweaking to check for that, but it won't be too hard.

kernel/dev/tty/tty.cpp
Kfeavel: That's fair. Is end already a label or will I need to update all of the returns to `goto end`?

kernel/dev/tty/tty.cpp
Kfeavel: In what cases should putchar return EOF? putchar is supposed to return the character printed, so do we return EOF on null / `\n`, or in what cases?

kernel/main.cpp
Kfeavel: Fair enough. I'll remove it and the makefile change.

Kfeavel:
At this point I'm pretty much ready to merge except I need @micahswitzer to explain how he would like the `goto` code to be implemented (since he already has an idea in mind) and in what cases `EOF` should be returned.

kernel/dev/tty/tty.cpp
micahswitzer: You'll need to add it.

kernel/dev/tty/tty.cpp
micahswitzer: I have no idea, I saw that you were depending on `putchar` returning `EOF` so I was clarifying what the expected behavior should be.

kernel/dev/tty/tty.cpp
micahswitzer: It looks like it will already behave that way

PR #125, opened 2020-09-13
Resolve: "RS232 Serial Input"

### LINKED ISSUE #71: RS232 Serial Input
RS232 Serial output works, but input is still unimplemented.

Kfeavel:
Closes #71


kernel/dev/serial/rs232.cpp
```diff
@@ -82,3 +90,14 @@ void px_rs232_init(uint16_t com_id) {
...
 }
+
+px_ring_buff_t* px_rs232_init_buffer(int size) {
+    // Allocate space for the input buffer
+    read_buffer = (px_ring_buff_t*)malloc(sizeof(px_ring_buff_t));
+    // Initialize the ring buffer
+    if (px_ring_buffer_init(read_buffer, size) == 0) {
+        return read_buffer;
+    } else {
+        return NULL;
```
micahswitzer, line 104: No error message for failed initialization?

kernel/dev/serial/rs232.cpp
```diff
@@ -48,11 +49,18 @@ void px_rs232_print(char* str) {
...
+    // Change carriage returns to newlines
+    if (in == '\r') {
+        in = '\n';
     }
+    // Create a string and print it so that the
+    // user can see what they're typing.
+    char str[2] = {in, '\0'};
     px_rs232_print(str);
+    // Add the character to the circular buffer
+    px_ring_buffer_enqueue(read_buffer, (uint8_t)in);
```
micahswitzer, line 66: What happens if someone sends serial data before the buffer is initialized? Do we just wait to enable interrupts until after both have been initialized?

kernel/dev/serial/rs232.cpp
```diff
@@ -82,3 +90,14 @@ void px_rs232_init(uint16_t com_id) {
         "Panix Serial Output Debugger\n\n"
     );
 }
+
+px_ring_buff_t* px_rs232_init_buffer(int size) {
```
micahswitzer, line 97: Does this function ever get called? I don't seen any references to it.

sysroot/usr/include/kernel/dev/serial/rs232.hpp
```diff
@@ -44,4 +45,14 @@ void px_rs232_init(uint16_t com_id);
...
+
+/**
+ * @brief Initializes the serial input buffer and
+ * returns a pointer to said buffer if successful.
+ * 
+ * @param size Buffer capacity (in bytes)
+ * @return px_ring_buff_t* Pointer to buffer. NULL
+ * if memory could not be allocated.
+ */
+px_ring_buff_t* px_rs232_init_buffer(int size);
```
micahswitzer, line 58: I feel weird about the serial subsystem returning a reference to it's internal data structure. Perhaps we should write an abstraction layer over the circular buffer?

kernel/dev/serial/rs232.cpp
Kfeavel: The appropriate errno would be set by the ring buffer's init function so no need to duplicate work, right?

kernel/dev/serial/rs232.cpp
Kfeavel: I'm fairly certain the input is just ignored.

kernel/dev/serial/rs232.cpp
Kfeavel: It's called in main. I didn't commit main because I didn't know if we actually wanted to use this yet since all it would do right now is print to screen. I figured I would wait until we had a real use for it.

sysroot/usr/include/kernel/dev/serial/rs232.hpp
Kfeavel: If you would like to do that then you can, but it would basically be creating duplicate functions that already exist in the ring buffer library, so I figured it would be easier to just use those with the returned pointer.

kernel/dev/serial/rs232.cpp
micahswitzer: Ah, right.

kernel/dev/serial/rs232.cpp
micahswitzer: So no null dereference? That's what it looks like would happen.

sysroot/usr/include/kernel/dev/serial/rs232.hpp
micahswitzer: Right, but this way we can control how the user interacts with the serial library. I know it feels like duplicated code/functions, but such is the pain of good abstractions.

kernel/dev/serial/rs232.cpp
micahswitzer: Ah okay, so this branch doesn't technically support input then?

kernel/dev/serial/rs232.cpp
Kfeavel: This code is pretty similar to what it was before, and we didn't get null dereferences then, but I can double check that later.

kernel/dev/serial/rs232.cpp
Kfeavel: It contains all of the logic and functions necessary to read input. I tested it on my machine and the input successfully printed to the screen. It does not, however, enable that functionality because I wasn't 100% of how we wanted to use it right now and figured we could decide that later.

sysroot/usr/include/kernel/dev/serial/rs232.hpp
Kfeavel: What functions would you like to be added then?

kernel/dev/serial/rs232.cpp
micahswitzer: I guess the ring buffer code should be checking that. It just doesn't feel right to me.

kernel/dev/serial/rs232.cpp
micahswitzer: Okay. I guess I just figured we'd enable it, and then let applications consume that information at a later time.

Kfeavel: I can do that. I'll make another commit later.

sysroot/usr/include/kernel/dev/serial/rs232.hpp
Kfeavel: I added get char, get string, and close pipe functions.

kernel/dev/serial/rs232.cpp
Kfeavel: Added to main.

Kfeavel:
Ready for another review.

kernel/dev/serial/rs232.cpp
```diff
@@ -101,3 +101,44 @@ px_ring_buff_t* px_rs232_init_buffer(int size) {
...
+    char* str = (char*)malloc(sizeof(char) * 1024);
+    // Keep reading until the buffer is empty or
+    // a newline is read.
+    while (!px_ring_buffer_is_empty(read_buffer)) {
+        uint8_t byte;
+        px_ring_buffer_dequeue(read_buffer, &byte);
+        str[idx] = (char)byte;
+        ++idx;
+        // Break if it's a newline
+        if ((char)byte == '\n') {
```
micahswitzer: The documentation states that it will stop on a null, but it looks like it will continue to consume data past a null.

kernel/dev/serial/rs232.cpp
```diff
@@ -101,3 +101,44 @@ px_ring_buff_t* px_rs232_init_buffer(int size) {
...
+char px_rs232_get_char() {
+    // Grab the last byte and convert to a char
+    uint8_t data;
+    px_ring_buffer_dequeue(read_buffer, &data);
+    return (char)data;
+}
+
+char* px_rs232_get_str() {
+    int idx = 0;
+    char* str = (char*)malloc(sizeof(char) * 1024);
```
micahswitzer: This is hard-coded to handle a maximum buffer size of 1024 bytes, but it could be larger. Maybe we let the user pass in their own buffer? I'm not a fan of `malloc`ing 1024 bytes for every call regardless.

kernel/dev/serial/rs232.cpp
```diff
@@ -101,3 +101,44 @@ px_ring_buff_t* px_rs232_init_buffer(int size) {
...
+        px_ring_buffer_dequeue(read_buffer, &byte);
+        str[idx] = (char)byte;
+        ++idx;
+        // Break if it's a newline
+        if ((char)byte == '\n') {
+            break;
+        }
+    }
+    // Add the null terminator
+    str[idx] = '\0';
```
micahswitzer: If the buffer was full, then this will write past the end of the array.

kernel/dev/serial/rs232.cpp
```diff
@@ -101,3 +101,44 @@ px_ring_buff_t* px_rs232_init_buffer(int size) {
...
+    // Return the string
+    return str;
+}
+
+int px_rs232_close() {
+    int ret = -1;
+    if (read_buffer != NULL) {
+        ret = px_ring_buffer_destroy(read_buffer);
```
micahswitzer, line 150: Should we also set `read_buffer` to `null` since it no longer is useful? Or does it need to be freed first?

micahswitzer (changes requested):

kernel/dev/serial/rs232.cpp
Kfeavel: It gets freed in the destroy call. We can just set it to NULL

kernel/dev/serial/rs232.cpp
Kfeavel: Resolved in the latest commit (hopefully)

kernel/dev/serial/rs232.cpp
Kfeavel: Resolved in the latest commit (hopefully)

kernel/dev/serial/rs232.cpp
Kfeavel: Resolved in the latest commit (hopefully)

kernel/dev/serial/rs232.cpp
Kfeavel: Resolved in the latest commit (hopefully)

Kfeavel:
Ready for re-re-review

kernel/dev/serial/rs232.cpp
```diff
@@ -82,3 +90,59 @@ void px_rs232_init(uint16_t com_id) {
...
+px_ring_buff_t* px_rs232_get_buffer() {
+    // Can return NULL. This is documented.
+    return read_buffer;
+}
+
+char px_rs232_get_char() {
+    // Grab the last byte and convert to a char
+    uint8_t data;
+    px_ring_buffer_dequeue(read_buffer, &data);
+    return (char)data;
```
micahswitzer, line 120: Couldn't this be undefined if the buffer is empty?

micahswitzer: So how do we know if we received a `null` vs. nothing at all?

Kfeavel:
At this point it would probably be best for you to finish the changes since you have a better sense of what an API should look like than I do.

kernel/dev/serial/rs232.cpp
Kfeavel: Added a note in the latest commit. We'll come back to this at a later date when we need better data transmission.

micahswitzer (approved):

PR #131, opened 2020-10-03
Change Compiler to Clang++

Kfeavel:
Some changes were made in order to get the code to compile since Clang forces a code to conform to the ISO standard a bit more (which is fine by me).

`-Wno-unused-function` was added so that we don't have to worry about unused functions for now since we may end up using them in the future. Other incompatible flags were also removed.


Kfeavel:
I'm so happy.

micahswitzer (approved):

PR #135, opened 2020-10-06
Add Debug Build Target

Kfeavel:


Kfeavel:
I don't know where else we would put more flags for now, so I guess this is ready to merge.

micahswitzer (commented):
Maybe we should consider adding a `px_debug_print` function/macro to avoid having to place `#ifdef`s everywhere.

micahswitzer (approved):

PR #150, opened 2020-11-28
Build System Redesign

Kfeavel:
Still going to stick with Make since it provides the greatest level of control, but we should figure out how we want to better integrate source files as our project starts to include libraries, other programs (when we get ELF loading), etc.


kernel/sys/early_panic.cpp
```diff
@@ -12,9 +12,12 @@
 #include <arch/arch.hpp>
 #include <dev/tty/tty.hpp>
 
+// Provide a function prototype to make the compiler warnings happy. Don't
+// want to make it public though, so we won't put it in the panic header.
+extern "C" void early_panic(const char *str);
+
 extern "C" void
 __attribute__ ((section(".early_text")))
-__attribute__ ((optnone))
```
micahswitzer, line 17: Why was this removed? Won't that break this functionality?

Makefile
```diff
@@ -41,26 +42,37 @@ QEMU = $(shell command -v qemu-system-$(QEMU_ARCH))
...
+	-Wshadow             \
+	-Wcast-align         \
+	-Wno-long-long       \
+	-Wpointer-arith      \
+	-Wconversion         \
+	-Wwrite-strings      \
+	-Wredundant-decls    \
+	-Wmissing-declarations
+# TODO: Include -Werror once we
+# Fix all of the warnings
```
micahswitzer: Have we not fixed all the warnings?

Makefile
```diff
@@ -102,10 +129,15 @@ OBJ_DIRS = $(subst kernel, obj, $(shell find kernel -type d))
 mkdir_obj_dirs:
 	mkdir -p $(OBJ_DIRS)
 
+# C source -> object
+obj/%.o: kernel/%.c $(HEADERS)
+	$(MAKE) mkdir_obj_dirs
+	$(CXX) $(CFLAGS) $(CPPFLAGS) -c -o $@ $<
```
micahswitzer: Shouldn't it be `$(CC)` for C code?

kernel/lib/printf.cpp
```diff
@@ -288,6 +294,10 @@ OK, I found my mistake. The math here is _always_ unsigned */
             default:
                 break;
             }
+            // Tell the compiler that this isn't an issue
+            // (at least I'm assuming it isn't)
+            [[fallthrough]];
```
micahswitzer, line 299: Maybe explain to the person looking at the code why this is the case as well

kernel/lib/semaphore.cpp
```diff
@@ -131,6 +132,8 @@ int px_sem_post(px_sem_t *sem) {
 }
 
 uint32_t px_sem_getval(px_sem_t *sem, uint32_t *val) {
+    // FIXME: I don't think val should be unused?
+    (void)val;
```
micahswitzer, line 136: Probably not. Can you make an issue for this?

kernel/mem/paging.cpp
```diff
@@ -88,7 +94,7 @@ static void px_paging_init_dir() {
         px_map_kernel_page_table(i, &page_tables[i]);
         // clear out the page tables
         for (int j = 0; j < PAGE_ENTRIES; j++) {
-            page_tables[i].pages[j] = (px_page_table_entry_t){ 0 };
+            page_tables[i].pages[j] = (px_page_table_entry_t){ /* ZERO */ };
```
micahswitzer, line 97: I prefer the actual zero, but maybe you did this because of warnings?

kernel/sys/early_panic.cpp
Kfeavel: GCC complained about it. Apparently it's a clang only thing.

Makefile
Kfeavel: Nope, because we added more with these new flags. Most of the ones left are due to implicit casting.

Makefile
Kfeavel: Fixed in the next commit.

kernel/lib/printf.cpp
Kfeavel: I'm not certain but I think it's because we want the code below to execute if it falls through so that we continue to print. If you have a better guess / comment just leave it here and I'll add it.

kernel/lib/semaphore.cpp
Kfeavel: #151 is now open

kernel/mem/paging.cpp
Kfeavel: Yes, I did this because of the warnings. In C this would set everything to 0, but in C++ this only sets the first struct member to 0. (The rest are technically set to 0, but the compiler complains since the correct way is to leave the `{ }` blank.

kernel/sys/early_panic.cpp
micahswitzer: Okay so we'll need to add the appropriate attribute for GCC. I believe it is something like `__attribute__((-O0))`.

kernel/mem/paging.cpp
micahswitzer: Ooo. I did not know that. I kinda don't like it, but I guess it is how it is. I'm also not a fan of the comment, but that's whatever.

Kfeavel: I can remove the comments, but I figured it would be better to have something explaining that it's all zero'd out.

kernel/lib/printf.cpp
micahswitzer: Ah okay yeah. It's kinda a horribly structured function in the first place.

kernel/sys/early_panic.cpp
Kfeavel: `__attribute__((-O0))` doesn't work. I'll see if I can't find what it is.

Makefile
micahswitzer: Is there any reason they can't be addressed now?

kernel/sys/early_panic.cpp
Kfeavel: Is `__attribute__((optimize("O0")))` what you're looking for?

Makefile
Kfeavel: I'm working on it. I've already fixed the 90% of other issues but type casting gets tricky, especially when we're dealing with structs which only use 20 bits of an int and stuff. (Yes, it does complain about that).

kernel/lib/printf.cpp
Kfeavel: Yeah, it is, but I don't feel like optimizing it right now.

kernel/sys/early_panic.cpp
micahswitzer: Probs

kernel/lib/printf.cpp
micahswitzer: All good

kernel/mem/paging.cpp
micahswitzer: Ok

kernel/sys/early_panic.cpp
Kfeavel: Yeah, it is. Found a stackoverflow post about it.

Makefile
Kfeavel: For right now `-Wconversions` has been removed since that was the only flag that was giving us trouble, and I'm not sure if it's something we want to keep. If it is then it can be added back, but we'll have to fix a **ton** of casting warnings

micahswitzer (approved):

PR #164, opened 2020-12-12
Critical Region Mutual Exclusion

### LINKED ISSUE #160: Critical Region Mutexing
Add mutexes / semaphores to critical areas of kernel

- [x] TTY / VGA Printing
- [x] Memory allocation / freeing
- [x] Serial IO

Kfeavel:
Closes #160

### DIFF
--- a/kernel/dev/serial/rs232.cpp
+++ b/kernel/dev/serial/rs232.cpp
@@ -20,8 +20,7 @@
 static px_ring_buff_t* read_buffer = NULL;
 uint8_t rs_232_line_index;
 uint16_t rs_232_port_base;
-px_mutex_t mutex_data;
-px_mutex_t mutex_init;
+px_mutex_t mutex_rs232;
 
 static int px_rs232_received();
 static int px_rs232_is_transmit_empty();
@@ -38,17 +37,17 @@ static int px_rs232_is_transmit_empty() {
 }
 
 static char px_rs232_read_char() {
-    px_mutex_lock(&mutex_data);
+    px_mutex_lock(&mutex_rs232);
     while (px_rs232_received() == 0);
-    px_mutex_unlock(&mutex_data);
+    px_mutex_unlock(&mutex_rs232);
     return px_read_byte(rs_232_port_base + RS_232_DATA_REG);
 }
 
 static void px_rs232_write_char(char c) {
-    px_mutex_lock(&mutex_data);
+    px_mutex_lock(&mutex_rs232);
     while (px_rs232_is_transmit_empty() == 0);
     px_write_byte(rs_232_port_base + RS_232_DATA_REG, c);
-    px_mutex_unlock(&mutex_data);
+    px_mutex_unlock(&mutex_rs232);
 }
 
 void px_rs232_print(const char* str) {
@@ -102,15 +101,15 @@ void px_rs232_init(uint16_t com_id) {
 }
 
 px_ring_buff_t* px_rs232_init_buffer(int size) {
-    px_mutex_lock(&mutex_init);
+    px_mutex_lock(&mutex_rs232);
     // Allocate space for the input buffer
     read_buffer = (px_ring_buff_t*)malloc(sizeof(px_ring_buff_t));
     // Initialize the ring buffer
     if (px_ring_buffer_init(read_buffer, size) == 0) {
-        px_mutex_unlock(&mutex_init);
+        px_mutex_unlock(&mutex_rs232);
         return read_buffer;
     } else {
-        px_mutex_unlock(&mutex_init);
+        px_mutex_unlock(&mutex_rs232);
         return NULL;
     }
 }
@@ -121,19 +120,19 @@ px_ring_buff_t* px_rs232_get_buffer() {
 }
 
 char px_rs232_get_char() {
-    px_mutex_lock(&mutex_data);
+    px_mutex_lock(&mutex_rs232);
     // Grab the last byte and convert to a char
     uint8_t data = 0;
     if (read_buffer != NULL)
     {
         px_ring_buffer_dequeue(read_buffer, &data);
     }
-    px_mutex_unlock(&mutex_data);
+    px_mutex_unlock(&mutex_rs232);
     return (char)data;
 }
 
 int px_rs232_get_str(char* str, int max) {
-    px_mutex_lock(&mutex_data);
+    px_mutex_lock(&mutex_rs232);
     int idx = 0;
     // Keep reading until the buffer is empty or
     // a newline is read.
@@ -153,18 +152,18 @@ int px_rs232_get_str(char* str, int max) {
             break;
         }
     }
-    px_mutex_unlock(&mutex_data);
+    px_mutex_unlock(&mutex_rs232);
     // Return the string
     return idx;
 }
 
 int px_rs232_close() {
-    px_mutex_lock(&mutex_init);
+    px_mutex_lock(&mutex_rs232);
     int ret = -1;
     if (read_buffer != NULL) {
         ret = px_ring_buffer_destroy(read_buffer);
         read_buffer = NULL;
     }
-    px_mutex_unlock(&mutex_init);
+    px_mutex_unlock(&mutex_rs232);
     return ret;
 }
--- a/kernel/dev/serial/rs232.cpp
+++ b/kernel/dev/serial/rs232.cpp
@@ -15,10 +15,13 @@
 #include <arch/arch.hpp>
 #include <dev/serial/rs232.hpp>
 #include <mem/heap.hpp>
+#include <lib/mutex.hpp>
 
 static px_ring_buff_t* read_buffer = NULL;
 uint8_t rs_232_line_index;
 uint16_t rs_232_port_base;
+px_mutex_t mutex_data;
+px_mutex_t mutex_init;
 
 static int px_rs232_received();
 static int px_rs232_is_transmit_empty();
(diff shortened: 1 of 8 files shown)


kernel/dev/serial/rs232.cpp
```diff
@@ -35,13 +38,21 @@ static int px_rs232_is_transmit_empty() {
 }
 
 static char px_rs232_read_char() {
+    // Gain mutual exclusion
```
micahswitzer: Probably don't need all these comments. Readers of the code should understand what a mutex does.

kernel/dev/serial/rs232.cpp
```diff
@@ -67,6 +78,8 @@ static void px_rs232_callback(registers_t *regs) {
 }
 
 void px_rs232_init(uint16_t com_id) {
+    // Gain mutual exclusion
+    px_mutex_lock(&mutex_init);
```
micahswitzer: A lock shouldn't be necessary here since initialization is performed before we enable multitasking.

micahswitzer (commented):
I think we might want to re-consider how we structure the serial code especially since we want to start using drivers at some point soon. That's not an issue with what you wrote, it's just that the locking might make more sense once we do that.

kernel/dev/serial/rs232.cpp
Kfeavel: Resolved in 111248a

kernel/dev/serial/rs232.cpp
Kfeavel: Resolved in 111248a

Kfeavel:
@micahswitzer Ready for re-review.

kernel/dev/serial/rs232.cpp
```diff
@@ -15,10 +15,13 @@
 #include <arch/arch.hpp>
 #include <dev/serial/rs232.hpp>
 #include <mem/heap.hpp>
+#include <lib/mutex.hpp>
 
 static px_ring_buff_t* read_buffer = NULL;
 uint8_t rs_232_line_index;
 uint16_t rs_232_port_base;
+px_mutex_t mutex_data;
+px_mutex_t mutex_init;
```
micahswitzer: I'm concerned that we're not initializing these properly. I think the default of `0` will work for now, but we might want to change that in the future which would break this. Maybe we could create a macro that would declare and initialize these structures properly?

```c++
// in mutex.hpp
#define PX_MUTEX(name) px_mutex_t name = {}

// in .cpp files where we use mutexes
PX_MUTEX(mutex_data);
```

I'm open to other ideas as well.

Kfeavel: Yeah, we could do something similar to PTHREAD_MUTEX_INITIALIZER

jamesaorson: Why use a macro when you define this in a constructor?

A macro seems like both overkill and a way of subverting a final solution if we are just trying to abstract away initialization.

jamesaorson:
Seeing all the lock and unlock statements got me thinking. Panix could benefit from having a java-like `synchronized` block. We can do this with macros like [this post](https://stackoverflow.com/a/26177785) suggests.

kernel/dev/serial/rs232.cpp
micahswitzer: Very true. We could actually use C++ constructors to make initialization completely transparent.

jamesaorson: Yeah, let's not fall into the same trap of the Linux kernel by depending on C-era macros that are mimicking C++ features.

Kfeavel: Very true. I keep forgetting that Panix is our opportunity to break the mold. We can always add in compatibility later if desired, but for now we can do things differently / better.

Kfeavel: Default constructors have been added in 533581f
Pinging @JamesOrson & @micahswitzer for their input.

sysroot/usr/include/kernel/lib/mutex.hpp
```diff
@@ -14,6 +14,9 @@
 
 typedef struct px_mutex {
     bool locked;
+    px_mutex() {
+        locked = false;
+    };
```
jamesaorson: Also as an implementation thing, I think we should define as little as possible in header files.

The constructor declaration should be in the header and the implementation in the cpp file.

jamesaorson (changes requested):

sysroot/usr/include/kernel/lib/mutex.hpp
Kfeavel: I honestly have no clue how to do that with structs. They're basically classes, so should I just declare `px_mutex();` in the header and move everything else, struct and all into the implementation file?

jamesaorson: The only difference in C++ between structs and classes is this:
- `class` members are `private` by default
- `struct` members are `public` by default
They are exactly the same otherwise.

jamesaorson: The resulting header code would look like this
```c++
typedef struct px_mutex {
    bool locked;
    px_mutex();
} px_mutex_t;
```

Kfeavel: Okay, in the header I have
```C++
typedef struct px_mutex {
    bool locked;
    px_mutex();
} px_mutex_t;
```
and in the source file I have
```C++
px_mutex::px_mutex() {
    locked = false;
};
```
Does that seem right?

jamesaorson: > Okay, in the header I have
> 
> ```c++
> typedef struct px_mutex {
>     bool locked;
>     px_mutex();
> } px_mutex_t;
> ```
> 
> and in the source file I have
> 
> ```c++
> px_mutex::px_mutex() {
>     locked = false;
> };
> ```
> 
> Does that seem right?

Perfect!

jamesaorson (approved):

sysroot/usr/include/kernel/lib/mutex.hpp
Kfeavel: I haven't committed that yet so I'll go ahead and do that for both the mutexes and semaphores.

Kfeavel: Changed in cd4204d

jamesaorson (dismissed):

kernel/dev/serial/rs232.cpp
```diff
@@ -15,10 +15,13 @@
 #include <arch/arch.hpp>
 #include <dev/serial/rs232.hpp>
 #include <mem/heap.hpp>
+#include <lib/mutex.hpp>
 
 static px_ring_buff_t* read_buffer = NULL;
 uint8_t rs_232_line_index;
 uint16_t rs_232_port_base;
+px_mutex_t mutex_data;
```
micahswitzer: I think we need one mutex for both data and initialization.

kernel/mem/paging.cpp
```diff
@@ -14,12 +14,14 @@
...
 #include <lib/bitmap.hpp>
 #include <lib/stdio.hpp>
+#include <lib/mutex.hpp>
 #include <dev/serial/rs232.hpp>
 #include <stddef.h>
 
 #define KADDR_TO_PHYS(addr) ((addr) - KERNEL_BASE)
 
 static uint32_t machine_page_count;
+static px_mutex_t mutex_paging;
```
micahswitzer, line 24: Can you double-check that these global constructors are getting called? I'm not sure we ever made sure that was happening.

Kfeavel:
> Seeing all the lock and unlock statements got me thinking. Panix could benefit from having a java-like `synchronized` block. We can do this with macros like [this post](https://stackoverflow.com/a/26177785) suggests.

This would be something to consider. We can make a discussion for it and talk about whether or not it would be something worth adding / how to go about adding it.

kernel/dev/serial/rs232.cpp
Kfeavel: Fixed in 285ec66

kernel/mem/paging.cpp
Kfeavel: The global constructor is *not* getting called. I put a print statement in there and it didn't do anything, plus we never actually make a call to it in `boot.s` or in our C/C++ code anywhere.

JamesOrson: Is this something we really need to check? This is how you construct an object on the heap in C++

Kfeavel:
Fixed the merge conflict. Should be ready for re-review.

Kfeavel:
Oh wait, we should probably fix the global constructor issue before continuing. @micahswitzer do you have any insight into that issue?

micahswitzer:
> 
> 
> Oh wait, we should probably fix the global constructor issue before continuing. @micahswitzer do you have any insight into that issue?

Not really. I know that we need to define a section where the global constructors will be placed, and then we need to iterate through them and call each one. I'm pretty sure the proper place to do that is before you call `main`, but I think you can get away with it being *the first* thing you do in `main`.

I can look into this further if you would like. We definitely need that working before we merge this code in.

kernel/mem/paging.cpp
micahswitzer: #172 has been merged.

micahswitzer:
The constructor issues have been resolved. Please rebase on `dev`.

Kfeavel:
@micahswitzer Rebased on top of dev. Ready for review.

micahswitzer (approved):
Looks good. Maybe we should consider naming C++ structs in pascal case to distinguish them from C-like structs?

PR #167, opened 2020-12-17
Add macros to check mutex validity and acquire a lock

### LINKED ISSUE #166: Reduce repetitive code and add constants in mutex.cpp
The `mutex.cpp` implementation file has a lot of simple repetitive code.

Two prime examples are:

1. ```c++
    if (mutex == NULL)
    {
        errno = EINVAL;
        return -1;
    }
2. ```c++
    if (__atomic_test_and_set(&mutex->locked, __ATOMIC_RELEASE))
    {
        errno = EINVAL;
        return -1;
    }

Also, throughout the file there are a lot of "magic numbers". There is no reason to use raw `int` return types when we could use something like
```c++
enum class MutexResult {
    success = 0,
    error = -1,
};
```

jamesaorson:
Closes #166

### DIFF
--- a/kernel/lib/mutex.cpp
+++ b/kernel/lib/mutex.cpp
@@ -23,13 +23,18 @@
 #include <mem/heap.hpp>
 #include <stddef.h>
 
+#define ACQUIRE_MUTEX_LOCK(mutex) __atomic_test_and_set(&mutex->locked, __ATOMIC_RELEASE)
+
+#define IS_MUTEX_VALID(mutex) { \
+    if (mutex == NULL)          \
+    {                           \
+        errno = EINVAL;         \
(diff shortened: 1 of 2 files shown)


Kfeavel (approved):
Looks good to me. Micah loves macros so he'll be happy about this change.

micahswitzer (approved):
I love it! Can we do the same for the semaphores?

jamesaorson:
> Looks good to me. Micah loves macros so he'll be happy about this change.

Macors are as

> I love it! Can we do the same for the semaphores?

I will add a second commit with the semaphores.

Kfeavel (approved):
Looks good to me.

PR #170, opened 2020-12-18
WIP: Basic Support for C++ & Basic Queues

Kfeavel:


sysroot/usr/include/kernel/lib/array_queue.hpp
```diff
@@ -0,0 +1,68 @@
...
+}
+
+template<typename T>
+void ArrayQueue<T>::enqueue(T i)
+{
+    if (size() == capacity())
+        PANIC("Attempt to fill queue past it's capacity");
+    m_items[m_head] = i;
+    m_head = (m_head + 1) % ARRAY_QUEUE_CAPACITY;
+    m_size++;
```
jamesaorson, line 50: Why don't we ever expand the internal array once we fill it?

jamesaorson (changes requested):

sysroot/usr/include/kernel/lib/array_queue.hpp
micahswitzer: This was designed to be a "quick and dirty" implementation to get a different feature working and I never ended up needing it. It was not designed to use dynamic memory allocation and therefore has no support for resizing. I'm hesitant to keep this code around since it is so limited. We should instead consider a resizable array queue and probably also a linked-list queue.

micahswitzer:
Closing this because it's a very limited implementation and we'll need to make something better for it to actually be useful.

PR #171, opened 2020-12-18
Multitasking Mutual Exclusion

### LINKED ISSUE #147: Multitasking Aware Mutexes & Semaphores

Kfeavel:
Closes #147

### DIFF
--- a/kernel/dev/serial/rs232.cpp
+++ b/kernel/dev/serial/rs232.cpp
@@ -60,9 +60,7 @@ static int vprintf_helper(unsigned c, void **ptr)
 int px_rs232_vprintf(const char* fmt, va_list args)
 {
     int retval;
-    //px_mutex_lock(&mutex_rs232);
     retval = px_do_printf(fmt, args, vprintf_helper, NULL);
-    //px_mutex_unlock(&mutex_rs232);
     return retval;
 }
 
@@ -78,11 +76,9 @@ int px_rs232_printf(const char *format, ...)
 }
 
 void px_rs232_print(const char* str) {
-    //px_mutex_lock(&mutex_rs232);
     for (int i = 0; str[i] != 0; i++) {
         px_rs232_write_char(str[i]);
     }
-    //px_mutex_unlock(&mutex_rs232);
 }
 
 static void px_rs232_callback(registers_t *regs) {
--- a/kernel/dev/serial/rs232.cpp
+++ b/kernel/dev/serial/rs232.cpp
@@ -16,11 +16,13 @@
 #include <dev/serial/rs232.hpp>
 #include <mem/heap.hpp>
 #include <lib/mutex.hpp>
+#include <stdarg.h>
+#include <lib/stdio.hpp>
 
 static px_ring_buff_t* read_buffer = NULL;
 uint8_t rs_232_line_index;
 uint16_t rs_232_port_base;
-px_mutex_t mutex_rs232;
+px_mutex_t mutex_rs232("rs232");
 
 static int px_rs232_received();
 static int px_rs232_is_transmit_empty();
@@ -43,17 +45,44 @@ static char px_rs232_read_char() {
     return px_read_byte(rs_232_port_base + RS_232_DATA_REG);
 }
 
-static void px_rs232_write_char(char c) {
-    px_mutex_lock(&mutex_rs232);
+static inline void px_rs232_write_char(char c) {
     while (px_rs232_is_transmit_empty() == 0);
     px_write_byte(rs_232_port_base + RS_232_DATA_REG, c);
-    px_mutex_unlock(&mutex_rs232);
+}
+
+static int vprintf_helper(unsigned c, void **ptr)
+{
+    (void) ptr;
+    px_rs232_write_char((char)c);
+    return 0;
+}
+
+int px_rs232_vprintf(const char* fmt, va_list args)
+{
+    int retval;
+    //px_mutex_lock(&mutex_rs232);
+    retval = px_do_printf(fmt, args, vprintf_helper, NULL);
+    //px_mutex_unlock(&mutex_rs232);
+    return retval;
+}
+
+int px_rs232_printf(const char *format, ...)
+{
+    va_list args;
+    int ret_val;
+
+    va_start(args, format);
+    ret_val = px_rs232_vprintf(format, args);
+    va_end(args);
+    return ret_val;
 }
 
 void px_rs232_print(const char* str) {
+    //px_mutex_lock(&mutex_rs232);
     for (int i = 0; str[i] != 0; i++) {
         px_rs232_write_char(str[i]);
     }
+    //px_mutex_unlock(&mutex_rs232);
 }
 
 static void px_rs232_callback(registers_t *regs) {
--- a/kernel/lib/mutex.cpp
+++ b/kernel/lib/mutex.cpp
@@ -33,8 +33,8 @@
     }                           \
 }
 
-px_mutex::px_mutex() {
-    locked = false;
+px_mutex::px_mutex(const char *name) : locked(false) {
+    task_sync.dbg_name = name;
 };
 
 int px_mutex_init(px_mutex_t *mutex) {
@@ -61,8 +61,6 @@ int px_mutex_lock(px_mutex_t *mutex) {
         // Block the current kernel task
         TASK_ONLY px_tasks_sync_block(&mutex->task_sync);
     }
-    // aquire the sync structure
-    TASK_ONLY px_tasks_sync_aquire(&mutex->task_sync);
     // Success, return 0
     return 0;
 }
@@ -81,8 +79,6 @@ int px_mutex_trylock(px_mutex_t *mutex) {
 
 int px_mutex_unlock(px_mutex_t *mutex) {
     IS_MUTEX_VALID(mutex);
-    // Release the lock if it is locked
-    TASK_ONLY px_tasks_sync_release(&mutex->task_sync);
     // Clear the lock
     __atomic_clear(&mutex->locked, __ATOMIC_RELEASE);
     TASK_ONLY px_tasks_sync_unblock(&mutex->task_sync);
--- a/kernel/lib/mutex.cpp
+++ b/kernel/lib/mutex.cpp
@@ -41,6 +41,7 @@ int px_mutex_init(px_mutex_t *mutex) {
     IS_MUTEX_VALID(mutex);
     // Initialize the value to false
     mutex->locked = false;
+    TASKS_SYNC_INIT(&mutex->task_sync);
     // Success, return 0
     return 0;
 }
@@ -57,7 +58,8 @@ int px_mutex_lock(px_mutex_t *mutex) {
     // Check if the mutex is unlocked
     while (ACQUIRE_MUTEX_LOCK(mutex))
     {
-        // Busy wait while trying to get the lock
+        // Block the current kernel task
+        px_tasks_block_current(TASK_BLOCKED);
     }
     // Success, return 0
     return 0;
--- a/kernel/lib/mutex.cpp
+++ b/kernel/lib/mutex.cpp
@@ -33,7 +33,9 @@
     }                           \
 }
 
-px_mutex::px_mutex(const char *name) : locked(false) {
+px_mutex::px_mutex(const char *name)
+    : locked(false)
+{
     task_sync.dbg_name = name;
 };
 
--- a/kernel/lib/mutex.cpp
+++ b/kernel/lib/mutex.cpp
@@ -41,7 +41,7 @@ int px_mutex_init(px_mutex_t *mutex) {
     IS_MUTEX_VALID(mutex);
     // Initialize the value to false
     mutex->locked = false;
-    TASKS_SYNC_INIT(&mutex->task_sync);
+    px_tasks_sync_init(&mutex->task_sync);
     // Success, return 0
     return 0;
 }
@@ -59,8 +59,10 @@ int px_mutex_lock(px_mutex_t *mutex) {
     while (ACQUIRE_MUTEX_LOCK(mutex))
     {
         // Block the current kernel task
-        px_tasks_block_current(TASK_BLOCKED);
+        TASK_ONLY px_tasks_sync_block(&mutex->task_sync);
     }
+    // aquire the sync structure
+    TASK_ONLY px_tasks_sync_aquire(&mutex->task_sync);
     // Success, return 0
     return 0;
 }
@@ -79,8 +81,11 @@ int px_mutex_trylock(px_mutex_t *mutex) {
 
 int px_mutex_unlock(px_mutex_t *mutex) {
     IS_MUTEX_VALID(mutex);
-    // Clear the lock if it is locked
+    // Release the lock if it is locked
+    TASK_ONLY px_tasks_sync_release(&mutex->task_sync);
+    // Clear the lock
     __atomic_clear(&mutex->locked, __ATOMIC_RELEASE);
+    TASK_ONLY px_tasks_sync_unblock(&mutex->task_sync);
     // Success, return 0
     return 0;
 }
--- a/kernel/lib/printf.cpp
+++ b/kernel/lib/printf.cpp
@@ -338,6 +338,7 @@ int px_ksprintf(char* buf, const char* fmt, ...)
 static int vprintf_help(unsigned c, void** ptr)
 {
     (void)ptr;
+    // we use the unlocked putchar here becaus we lock at the printf level
     putchar_unlocked((char)c);
     return 0;
 }
--- a/kernel/lib/printf.cpp
+++ b/kernel/lib/printf.cpp
@@ -91,19 +91,13 @@ Using & for division here, so STACK_WIDTH must be a power of 2. */
 2^32-1 in base 8 has 11 digits (add 5 for trailing NUL and for slop) */
 #define PR_BUFLEN 16
 
-typedef int (*fnptr_t)(unsigned c, void** helper);
-
-/* Function declarations */
-int do_printf(const char* fmt, va_list args, fnptr_t fn, void* ptr);
-int vprintf_help(unsigned c, void** ptr);
-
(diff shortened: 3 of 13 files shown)


Kfeavel:
Mutexes are done, semaphores are next.

sysroot/usr/include/kernel/lib/mutex.hpp
```diff
@@ -8,25 +8,29 @@
  * @copyright Copyright the Panix Contributors (c) 2020
  * 
  */
-
+#ifndef PANIX_MUTEX_HPP
+#define PANIX_MUTEX_HPP
```
jamesaorson: Why not use `#pragma once`?

sysroot/usr/include/kernel/lib/semaphore.hpp
```diff
@@ -8,13 +8,17 @@
  * @copyright Copyright the Panix Contributors (c) 2020
  * 
  */
+#ifndef PANIX_SEMAPHORE_HPP
+#define PANIX_SEMAPHORE_HPP
```
jamesaorson: Why not use `#pragma once`?

sysroot/usr/include/kernel/sys/tasks.hpp
```diff
@@ -12,13 +12,14 @@
 #ifndef PANIX_TASKS_HPP
 #define PANIX_TASKS_HPP
```
jamesaorson: Why not use `#pragma once`?

sysroot/usr/include/kernel/sys/tasks.hpp
```diff
@@ -37,26 +39,104 @@ struct px_task
...
+ * and task state struct. If the storage parameter is provided, the px_task_t struct
+ * provided will be written to. If NULL is passed as the storage parameter, a pointer
+ * to a allocated task will be returned.
+ *
+ * @param entry Task function entry point
+ * @param storage Task stack structure (if NULL, a pointer to the task is returned)
+ * @param state Task state structure
+ * @return px_task_t* Pointer to the created kernel task
+ */
+px_task_t *px_tasks_new(void (*entry)(void), px_task_t *storage, px_task_state state, const char *name);
```
jamesaorson, line 93: Why do we need this function when we could make them constructors on the `px_task_t` type? By using an external function to "initialize" the task, that means that we have to account for potential misuse of task objects before proper initialization.

sysroot/usr/include/kernel/sys/tasks.hpp
```diff
@@ -12,13 +12,14 @@
 #ifndef PANIX_TASKS_HPP
 #define PANIX_TASKS_HPP
 
-#include <stdint.h>    // Data type definitions
+#include <stdint.h>         // Data type definitions
 #include <arch/arch.hpp>    // Architecture specific features
 #include <mem/paging.hpp>
 
 #define TIME_SLICE_SIZE (1 * 1000 * 1000ULL)
```
jamesaorson, line 18: I have a question about all the functions in this file as a whole. Is there a reason we have defined these all as separate functions instead of as static functions and member functions of `px_task_t`?

All of these functions take at least one task parameter and have `px_tasks` prefixed to them, sort of cluttering up the code a bit imo and making it easier to have my eyes glaze over while reading the code.

jamesaorson (dismissed):

sysroot/usr/include/kernel/sys/tasks.hpp
micahswitzer: This has been a C-only project pretty much until now. I'm totally for using C++ objects and structs with member functions, but we also must be careful that we don't just end up re-writing the entire kernel (unless we want to do that). Maybe we can continue the discussion of incorporating more C++ code/features somewhere else?

sysroot/usr/include/kernel/sys/tasks.hpp
micahswitzer: I agree with this. See my other comment on using C++ features.

sysroot/usr/include/kernel/lib/mutex.hpp
micahswitzer: We didn't know about it when we started. I'm all for it now though.

Kfeavel: Personally, I prefer this style over pragma once, but I'll concede my preference if everyone else wants pragma.

jamesaorson: I don't think there is too great of an argument to use this style over `#pragma once` unless you are using it as documentation for what file you are in, but there should be a top of the file docstring doing so.

Generally, it seems the standard for modern code is to use `#pragma once` as it is less error-prone and pollutes the preprocessor scope less.
https://docs.microsoft.com/en-us/cpp/preprocessor/once?view=msvc-160
https://stackoverflow.com/questions/787533/is-pragma-once-a-safe-include-guard

Kfeavel: I didn't realize there were actual differences / benefits to using pragma once. If that's the case then I'm alright with using it.

Kfeavel: Apparently GCC optimizes include guards to act almost identically to pragma once, but we can still make the switch.

sysroot/usr/include/kernel/sys/tasks.hpp
Kfeavel: I personally would prefer to avoid OOP unless it makes sense. As an example, it makes sense for drivers to be OOP since they all need to follow a certain format. However, I prefer procedural programming over OOP and have tried to keep to that throughout Panix. The main reason that C++ was chosen was simply to have the choice to do OOP when it is desired without having to mix the two languages.

jamesaorson: I understand that we all have wounds from bad OOP. Same here.

But in all honesty, what's the difference in this case with moving the passed in task parameter to a `this` for the functions, while also removing the repetitive `px_tasks_` prefix? In a sense, by moving them to member functions, you get free namespacing and shorter function names.

jamesaorson: To lay my cards on the table though, I am very inclined to functional OOP. I hate state and I love types and classes. Hence my pushes for things like enums and functions defined within types.

sysroot/usr/include/kernel/lib/mutex.hpp
Kfeavel: See #174 and its associated MR. Resolved in 67712a8.

sysroot/usr/include/kernel/lib/semaphore.hpp
Kfeavel: See #174 and its associated MR. Resolved in 67712a8.

sysroot/usr/include/kernel/sys/tasks.hpp
Kfeavel: See #174 and its associated MR. Resolved in 67712a8.

sysroot/usr/include/kernel/sys/tasks.hpp
Kfeavel: My biggest problem with OOP is that I've been bitten by it hard before when doing kernel development. Problems with OOP and getting OOP to work properly is part of the reason Panix v2 died and I would like to avoid that as much as possible. OOP just brings extra assumptions and requirements into play that normally would not be a concern when doing procedural programming. I also find it easier to debug procedural code vs. OOP code since inheritance and other factors do not have to be taken into account.

Kfeavel: All future discussion on this issue should be held in #176. Once a conclusion has been determined a separate MR can be opened to address the issue.

sysroot/usr/include/kernel/sys/tasks.hpp
Kfeavel: All future discussion on this issue should be held in #176. Once a conclusion has been determined a separate MR can be opened to address the issue.

Kfeavel: And just for reference -- I think the struct constructors are a good idea. One less "moving part" that needs to be initialized.

sysroot/usr/include/kernel/sys/tasks.hpp
Kfeavel: I'm going to resolve this thread for now since we have an open discussion for this. If it is determined that OOP is the way to go, we should convert large chunks at a time in separate MRs. For now, I think it would be better to try to get this finished as soon as possible.

kernel/dev/serial/rs232.cpp
```diff
@@ -43,17 +45,44 @@ static char px_rs232_read_char() {
...
+{
+    (void) ptr;
+    px_rs232_write_char((char)c);
+    return 0;
+}
+
+int px_rs232_vprintf(const char* fmt, va_list args)
+{
+    int retval;
+    //px_mutex_lock(&mutex_rs232);
```
Kfeavel: Shouldn't these locks be removed if they're commented out?

kernel/dev/serial/rs232.cpp
```diff
@@ -43,17 +45,44 @@ static char px_rs232_read_char() {
...
+    int ret_val;
+
+    va_start(args, format);
+    ret_val = px_rs232_vprintf(format, args);
+    va_end(args);
+    return ret_val;
 }
 
 void px_rs232_print(const char* str) {
+    //px_mutex_lock(&mutex_rs232);
```
Kfeavel: See previous comment.

kernel/lib/mutex.cpp
```diff
@@ -33,8 +33,8 @@
     }                           \
 }
 
-px_mutex::px_mutex() {
-    locked = false;
+px_mutex::px_mutex(const char *name) : locked(false) {
```
Kfeavel: Personally I prefer all of the property setters to be on a new line so that it's obvious they are getting set.

kernel/lib/printf.cpp
```diff
@@ -341,17 +335,20 @@ int px_ksprintf(char* buf, const char* fmt, ...)
 }
 /*****************************************************************************
 *****************************************************************************/
-int vprintf_help(unsigned c, void** ptr)
+static int vprintf_help(unsigned c, void** ptr)
```
Kfeavel, line 338: Do we want to start prepending `_` to private functions?

kernel/lib/putchar.cpp
```diff
@@ -41,11 +41,21 @@ uint16_t px_ansi_vga_table[16] = {
...
+    int retval;
     // Gain mutual exclusion
     px_mutex_lock(&put_mutex);
+    retval = putchar_unlocked(c);
+    // Release mutual exclusion
+    px_mutex_unlock(&put_mutex);
+    return retval;
+}
+
+int putchar_unlocked(char c) {
```
Kfeavel, line 58: I get why you named this function `putchar_unlocked`, but I feel like that's a bad name given that there are locks around it in other functions (like `putchar`).

kernel/sys/tasks.cpp
```diff
@@ -98,6 +118,39 @@ static inline uint64_t _get_cpu_time_ns()
     return (__rdtsc()) / _instr_per_ns;
 }
 
+static void _print_task(const px_task_t *task)
+{
+    px_rs232_printf("%s is %s\n", task->name, _state_names[task->state]);
+}
+
+#ifdef DEBUG
+#define TASK_ACTION(action, task) do { px_rs232_print(action " "); _print_task(task); } while(0)
```
Kfeavel, line 127: Do we really need the `do { ... } while(0)` loop here?

kernel/sys/tasks.cpp
```diff
@@ -143,7 +197,11 @@ static void _task_starting()
     // it is run in the context of the new task
 
     // the task before this caused the scheduler to lock
-    _release_scheduler_lock();
+    // so we must unlock here
+    _scheduler_lock--;
```
Kfeavel, line 201: Why are you calling `--` on a lock? We don't have an operator overwritten for that, do we?

kernel/sys/tasks.cpp
```diff
@@ -338,6 +399,10 @@ static void _schedule()
     }
     // reset the time slice because a new task is being scheduled
     _time_slice_remaining = TIME_SLICE_SIZE;
+#ifdef DEBUG
+    px_rs232_print("switching to ");
```
Kfeavel, line 403: Shouldn't this be converted to the new `px_rs232_printf` function?

sysroot/usr/include/kernel/lib/stdio.hpp
```diff
@@ -11,11 +11,16 @@
 
 #include <stdint.h>
 #include <stdarg.h>
+#include <lib/mutex.hpp>
 
 #ifndef EOF
 #define EOF (-1)
 #endif
 
+typedef int (*fnptr_t)(unsigned c, void** helper);
```
Kfeavel, line 21: Shouldn't this be a `size_t` instead of an `int`?

kernel/sys/tasks.cpp
micahswitzer: Not really, but it's the "proper" thing to do. See https://stackoverflow.com/questions/154136/why-use-apparently-meaningless-do-while-and-if-else-statements-in-macros

kernel/sys/tasks.cpp
micahswitzer: Nope. It's just a plain integer. It's decremented because it's a recursive lock.

kernel/sys/tasks.cpp
micahswitzer: Yeah... but it's just debug code and it doesn't *need* formatting capabilities. Also, the regular print function is much faster than `printf`.

sysroot/usr/include/kernel/lib/stdio.hpp
micahswitzer: I just copied this from `stdio.cpp`, but I do believe `int` is correct. Why do you think it should be a `size_t`?

kernel/dev/serial/rs232.cpp
micahswitzer: Probably, but I left them in because I'd like to have some locking for regular tasks there, but to not have locking for scheduling debug code.

kernel/lib/mutex.cpp
micahswitzer: Okay, I can do that.

kernel/lib/printf.cpp
micahswitzer: We can. It was an experiment I started in the task/scheduling code that we can decide whether or not to carry over to everything else. Generally marking something as `static` is good enough, but I didn't want to get confused between the public and private versions of certain scheduling functions.

sysroot/usr/include/kernel/lib/stdio.hpp
Kfeavel: If it's a pointer then it should be the length of whatever `size_t` is, right?

kernel/lib/putchar.cpp
micahswitzer: Yeah I do see how that could be confusing, but I do feel like this is good name. I'm totally open to changing the name, but I can't think of something at the moment.

kernel/sys/tasks.cpp
Kfeavel: Fair enough.

kernel/sys/tasks.cpp
Kfeavel: Sounds good.

sysroot/usr/include/kernel/lib/stdio.hpp
micahswitzer: `int` is the return type of the function pointer.

kernel/sys/tasks.cpp
Kfeavel: I still don't like it but it makes sense.

kernel/lib/printf.cpp
Kfeavel: Sounds good. I'll make an issue for it.

kernel/dev/serial/rs232.cpp
Kfeavel: You could just do an `#ifndef DEBUG`

micahswitzer: But this function can be used in both places in the same build

micahswitzer: I'll just remove the comments for now

kernel/lib/putchar.cpp
Kfeavel: Would `putchar_internal` work? Or something like that?

sysroot/usr/include/kernel/lib/stdio.hpp
Kfeavel: Ah, right. That makes sense. Is `fnptr_t` actually used anywhere? If not, maybe we should just remove it?

kernel/lib/putchar.cpp
micahswitzer: Maybe, but it's not actually internal. I feel like unlocked is really the best name for it. Maybe I can just add some documentation to that function explaining the name?

sysroot/usr/include/kernel/lib/stdio.hpp
micahswitzer: It's used in `px_do_printf` on the line below.

kernel/lib/putchar.cpp
Kfeavel: Fair enough. I'll resolve it.

sysroot/usr/include/kernel/lib/stdio.hpp
Kfeavel: Alright then. I'm not a fan of typedef'ing it, but we'll leave it for now.

micahswitzer (approved):
I'm approving on behalf of @Kfeavel since he opened the PR even though I was its primary author.

PR #186, opened 2020-12-29
Resolve: "Set Panix Version in Makefile"

### LINKED ISSUE #185: Define Panix Version in Makefile
The current Panix version should be defined in the master Makefile and then set in `CFLAGS` just like the current commit.

Kfeavel:
Closes #185

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,11 @@ export PROJ_NAME = panix
 # Makefile flags
 # prevent make from showing "entering/leaving directory" messages
 export MAKEFLAGS   += --no-print-directory
-export GIT_VERSION := "$(shell git describe --abbrev=8 --dirty --always --tags)"
+# Panix Version
+export GIT_COMMIT := "$(shell git describe --abbrev=8 --dirty --always --tags)"
+export VER_MAJOR := "0"
+export VER_MINOR := "3"
+export VER_PATCH := "0"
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -14,7 +14,11 @@ export PROJ_NAME = panix
 # Makefile flags
 # prevent make from showing "entering/leaving directory" messages
 export MAKEFLAGS   += --no-print-directory
-export GIT_VERSION := "$(shell git describe --abbrev=8 --dirty --always --tags)"
+# Panix Version
+export GIT_COMMIT := "$(shell git describe --abbrev=8 --dirty --always --tags)"
+export VER_MAJOR := "0"
+export VER_MINOR := "3"
```
micahswitzer: Are we bumping version numbers right before, or right after release? It looks like we're currently bumping right before, but I do really like doing it immediately after.

micahswitzer (dismissed):

Makefile
Kfeavel: The current version is 0.3.0 and we'll bump it to 0.4.0 after this December release.

Kfeavel: Bumped.

micahswitzer: We're going to start bumping versions following releases.

micahswitzer (approved):

PR #192, opened 2021-01-02
Resolve: "Move Docker Container to Arch Linux"

### LINKED ISSUE #95: Move Docker Container to Arch Linux
Our current Docker container is a _**thicc**_ 1.87 GB because of ~Debian~ *the distro which shall not be named*, so we would like to force it to go on a diet and lose some that that thiccness. Alpine with GNU instead of MUSL seems to be the perfect replacement.

Kfeavel:
Closes #95

### DIFF
--- a/.scuba.yml
+++ b/.scuba.yml
@@ -1,1 +1,1 @@
-image: docker.pkg.github.com/panix-os-dev-team/panix/panixbuild:1.0.0
+image: docker.pkg.github.com/panix-os-dev-team/panix/panixbuild:3.0.0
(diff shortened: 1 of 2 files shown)


Kfeavel:
CodeFactor is not happy with the changes, but some of them have to be made in order to get Scuba to work on macOS. Waiting for input from @JamesOrson since he's better at Docker than anyone else on the team.

Kfeavel:
CodeFactor is complaining about the last user being root, but from what I can tell, scuba needs that to create users.

micahswitzer (approved):
Awesome work!

PR #196, opened 2021-02-22
WIP: #183: Prepend Underscore to All Private Functions / Methods

### LINKED ISSUE #183: Prepend Underscore to All Private Functions / Methods

Kfeavel:
Closes #183


Kfeavel:
I also added underscores to private variables, but if we don't like that then I can just drop the commit.

PR #200, opened 2021-03-02
Resolve: "Print Stivale2 Info on Boot"

### LINKED ISSUE #199: Print Stivale2 Info on Boot

Kfeavel:
Closes #199


kernel/arch/i386/boot/bootparse.cpp
```diff
@@ -116,10 +123,166 @@ void px_parse_multiboot2(void *info)
...
+            {
+                px_rs232_printf("Unknown Stivale2 tag: %d\n", tag->identifier);
+                break;
+            }
+        }
+
+        tag = (struct stivale2_tag*)tag->next;
+    }
+    px_rs232_printf("Done\n");
+}
```
micahswitzer: Newline missing

Kfeavel: Resolved in 02b0cf63de13e5ef1e9cb236f5c681bbe934d310

kernel/arch/i386/boot/bootparse.cpp
```diff
@@ -116,10 +123,166 @@ void px_parse_multiboot2(void *info)
...
+    static const char *mmap_types[] = {
+        [0] = "Invalid",
+        [STIVALE2_MMAP_USABLE] = "Available",
+        [STIVALE2_MMAP_RESERVED] = "Reserved",
+        [STIVALE2_MMAP_ACPI_RECLAIMABLE] = "ACPI reclaimable",
+        [STIVALE2_MMAP_ACPI_NVS] = "Non-volatile storage",
+        [STIVALE2_MMAP_BAD_MEMORY] = "Bad RAM"
+    };
+    // List of memory map entries
+    struct stivale2_mmap_entry* mmap_list = (struct stivale2_mmap_entry*)mmap->memmap;
```
jamesaorson: Be a bit easier to read using auto on the left hand side

kernel/arch/i386/boot/bootparse.cpp
```diff
@@ -116,10 +123,166 @@ void px_parse_multiboot2(void *info)
...
+    for (uint64_t i = 0; i < mmap->entries; i++)
+    {
+        struct stivale2_mmap_entry* entry = &mmap_list[i];
```
jamesaorson: Be a bit easier to read using auto on the left hand side

kernel/arch/i386/boot/bootparse.cpp
```diff
@@ -116,10 +123,166 @@ void px_parse_multiboot2(void *info)
...
+{
+    auto fixed = (struct stivale2_struct*)info;
+    for (uintptr_t page = ((uintptr_t)info & PAGE_ALIGN) + PAGE_SIZE;
+         page <= ((uintptr_t)info & PAGE_ALIGN);
+         page += PAGE_SIZE) {
+        px_rs232_printf("Mapping bootinfo at 0x%08x\n", page);
+        px_map_kernel_page(VADDR(page), page);
+    }
+    // Walk the list of tags in the header
+    struct stivale2_tag* tag = (struct stivale2_tag*)(fixed->tags);
```
jamesaorson: Be a bit easier to read using auto on the left hand side

kernel/arch/i386/boot/bootparse.cpp
```diff
@@ -116,10 +123,166 @@ void px_parse_multiboot2(void *info)
...
+                auto memmap = (struct stivale2_struct_tag_memmap*)tag;
+                px_rs232_printf("Stivale2 memory map:\n");
+                print_stivale2_mmap(memmap);
+                break;
+            }
+            case STIVALE2_STRUCT_TAG_FRAMEBUFFER_ID:
+            {
+                auto framebuffer = (struct stivale2_struct_tag_framebuffer*)tag;
+                px_rs232_printf("Stivale2 framebuffer:\n");
+                px_rs232_printf("  Address: 0x%08X", framebuffer->framebuffer_addr);
```
jamesaorson, line 208: Maybe the indented entries should use tab characters? Thoughts?

jamesaorson (changes requested):

kernel/arch/i386/boot/bootparse.cpp
Kfeavel: Currently our `printf` implementation doesn't support `\t` otherwise I absolutely would.

kernel/arch/i386/boot/bootparse.cpp
Kfeavel: Resolved in ced2fd7a2b501b80624e9e68ebf765e4d6644c52

kernel/arch/i386/boot/bootparse.cpp
Kfeavel: Resolved in ced2fd7a2b501b80624e9e68ebf765e4d6644c52

kernel/arch/i386/boot/bootparse.cpp
Kfeavel: Resolved in ced2fd7a2b501b80624e9e68ebf765e4d6644c52

kernel/arch/i386/boot/bootparse.cpp
Kfeavel: Actually, it likely would look fine given that it's printing over serial as opposed to the BIOS text mode... Any thoughts @micahswitzer?

micahswitzer: I mean, I personally just like the small indentation. It's purely a stylistic thing, and I don't anticipate this code providing much value once we get past the new memory-mapping code. Ultimately it should be whatever we think looks best for ourselves.

Kfeavel: Personally I prefer `\t` but like you said, it probably won't stick around for long, and we don't have support for `\t` right now. However, I did open #201 and planned to go back and replace all appropriate spaces with `\t` so long as everyone is okay with it.

kernel/arch/i386/boot/bootparse.cpp
```diff
@@ -116,10 +123,166 @@ void px_parse_multiboot2(void *info)
...
+    {
+        auto entry = &mmap_list[i];
+        px_rs232_printf("  addr: 0x%02x%08x, length: 0x%02x%08x, type: %s\n",
+            (uint32_t)(entry->base >> 32) & 0xff, (uint32_t)(entry->base & UINT32_MAX),
+            (uint32_t)(entry->length >> 32) & 0xff, (uint32_t)(entry->length & UINT32_MAX),
+            // Have to do this because the IDs aren't in an order that can be
+            // mapped into an array like Multiboot. The last two types start at
+            // 0x1000, so we have to do some ternary operator magic.
+            (entry->type < 6 ? mmap_types[entry->type]
+                             : (entry->type == STIVALE2_MMAP_BOOTLOADER_RECLAIMABLE ? "Bootloader"
```
micahswitzer: It might be better to just write a function that returns the correct string. The split between an array, and a ternary operator feels kinda weird.

kernel/arch/i386/boot/bootparse.cpp
```diff
@@ -116,10 +123,166 @@ void px_parse_multiboot2(void *info)
...
+                                                                                    : "Kernel & Modules")));
+    }
+}
+
+void px_parse_stivale2(void *info)
+{
+    auto fixed = (struct stivale2_struct*)info;
+    for (uintptr_t page = ((uintptr_t)info & PAGE_ALIGN) + PAGE_SIZE;
```
micahswitzer: This loop doesn't iterate anything and simply ends up mapping the page *after* the info structure. It can be removed entirely.

kernel/arch/i386/boot/bootparse.cpp
```diff
@@ -116,10 +123,166 @@ void px_parse_multiboot2(void *info)
...
+                px_rs232_printf("Stivale2 memory map:\n");
+                print_stivale2_mmap(memmap);
+                break;
+            }
+            case STIVALE2_STRUCT_TAG_FRAMEBUFFER_ID:
+            {
+                auto framebuffer = (struct stivale2_struct_tag_framebuffer*)tag;
+                px_rs232_printf("Stivale2 framebuffer:\n");
+                px_rs232_printf("  Address: 0x%08X", framebuffer->framebuffer_addr);
+                px_rs232_printf("  Resolution: %ix%ix%i\n",
```
micahswitzer, line 209: Any reason you chose `%i` over `%d`? I know they're the same, but I feel like `%d` is more common.

kernel/arch/i386/boot/bootparse.cpp
```diff
@@ -116,10 +123,166 @@ void px_parse_multiboot2(void *info)
...
+            case STIVALE2_STRUCT_TAG_FB_MTRR_ID:
+            {
+                px_rs232_printf("  Framebuffer has MTRR\n");
+                break;
+            }
+            case STIVALE2_STRUCT_TAG_MODULES_ID:
+            {
+                auto modules = (struct stivale2_struct_tag_modules*)tag;
+                for (uint64_t i = 0; i < modules->module_count; i++) {
+                    stivale2_module mod = modules->modules[i];
```
micahswitzer: Any reason you copy the module structure vs. referencing it in memory?

kernel/arch/i386/boot/bootparse.cpp
```diff
@@ -116,10 +123,166 @@ void px_parse_multiboot2(void *info)
...
+                        mod.begin,
+                        mod.end
+                    );
+                }
+                break;
+            }
+            case STIVALE2_STRUCT_TAG_RSDP_ID:
+            {
+                auto rsdp = (struct stivale2_struct_tag_rsdp*)tag;
+                px_rs232_printf("ACPI RSDP: %08X\n", rsdp->rsdp);
```
micahswitzer: This is a really small thing, but would it be possible to share logic with Multiboot and print the RSDP info?

boot/limine.cfg
```diff
@@ -1,6 +1,7 @@
 TIMEOUT=1
 GRAPHICS=no
-THEME_COLOURS=60000000;aa0000;00aaff;aa5500;0000aa;aa00aa;9076de;aaaaaa
+
```
micahswitzer, line 3: Any reason for the extra newline and slight change in theme when we don't enable graphics?

jamesaorson (dismissed):

kernel/arch/i386/boot/bootparse.cpp
Kfeavel: I use `%i` because it describes the type that it's printing. I'm aware that `%d` is the same (technically there are slight differences), but given that Panix has no defined style guide, and considering that `%i` has been used in various other places throughout the kernel, I don't feel as though there's a need to change it until a definite style guide has been either chosen or created.

kernel/arch/i386/boot/bootparse.cpp
Kfeavel: None other than it was simple to index the array. I will change it to a pointer.

kernel/arch/i386/boot/bootparse.cpp
Kfeavel: We're only given a `uint64_t` for the rsdp and not a full struct. Is it worth going through the effort of converting the `uint64_t` into a Multiboot2 compliant struct just to print information?

boot/limine.cfg
Kfeavel: Newline doesn't necessarily have a purpose. Color changes apply to the text and not just the graphics when the framebuffer is enabled.

kernel/arch/i386/boot/bootparse.cpp
Kfeavel: I know that in the grand scheme of things a function call isn't that much time, but that seems a bit overkill when like you said before, we're going to be replacing all of this eventually when we have a use for it. I personally think ternaries are great ways to inline selection between two options.

boot/limine.cfg
micahswitzer: Ah cool. I didn't notice it, but that's nice.

kernel/arch/i386/boot/bootparse.cpp
Kfeavel: Resolved in 4122625

kernel/arch/i386/boot/bootparse.cpp
Kfeavel: Resolved in 4122625

kernel/arch/i386/boot/bootparse.cpp
micahswitzer: But you're selecting between more than 2 values...

Kfeavel: I see. I forgot that the others were in the array. I can make a function that returns the type. I'll just make it a switch statement.

Kfeavel: Resolved in 5dfcb9d

kernel/arch/i386/boot/bootparse.cpp
micahswitzer: Lol savage

kernel/arch/i386/boot/bootparse.cpp
micahswitzer: It's a pointer? Just cast it to a `struct RSDPDescriptor *`.

Kfeavel: I wouldn't assume that they're structured the same, but I'll give it a shot.

Kfeavel: Surprisingly it worked. Resolved in 19a72e6.

kernel/arch/i386/boot/bootparse.cpp
```diff
@@ -116,10 +122,179 @@ void px_parse_multiboot2(void *info)
...
+ * |___/\__|_|\_/\__,_|_\___/___|
+ */
+
+static const char* stivale2_mmap_type_to_string(uint32_t type)
+{
+    const char* str = "Invalid";
+    switch (type)
+    {
+        case STIVALE2_MMAP_USABLE:
+            str = "Available";
```
micahswitzer: Again super picky, but you can just return the string directly and that cuts down the number of lines by a factor of 2.

Kfeavel: Resolved in 8a32ed6

micahswitzer (approved):

PR #204, opened 2021-03-06
Upload image artifact

micahswitzer:
Hopefully we can use this to allow running an image in a web browser with an image pulled directly from the build artifacts.


Kfeavel (approved):
Looks good to me.

PR #206, opened 2021-04-16
WIP: Better Memory Management

Kfeavel:
Merged `mmv2` into `better-memory`. The goal is to separate the virtual, physical, memory ranges, and memory manager into their respective classes. Having an object oriented memory manager has its pros and cons. If this isn't the direction we want to go, then we can open a thread in the Discussions section.

@micahswitzer will surely have some opinions on the matter of memory management, and @JamesOrson will likely have some opinions on OOP, so I'm expecting hopefully something good will come out of this.


Kfeavel:
This is going to be a real pain to rebase, especially since I didn't stick to the idea of just refactoring the boot process and started moving architecture stuff around. Maybe it's best to drop commits, keep what we want, and start over?

Kfeavel:
Closing this due to scope creep and a significant number of changes since this PR was opened. @the-grue will help resolve a lot of issues here with the `nasm` to `C` conversion he's working on. After that, the scope for this refactor will be divided into smaller components.

PR #208, opened 2021-04-18
Resolve "Additional Compilation Warnings"

### LINKED ISSUE #207: Additional Compilation Warnings
`-Wundef`, `-Wformat=2`, `-Wdouble-promotion`, and `-Wconversion`.

`-Wconversion` is tricky since we have fields with 20 bit variables, and GCC complains about those, so for now we should fix as many issues as possible.

Kfeavel:
Closes #207

### DIFF
--- a/kernel/arch/i386/boot/bootparse.cpp
+++ b/kernel/arch/i386/boot/bootparse.cpp
@@ -273,8 +273,8 @@ void px_parse_stivale2(void *info)
                 uint8_t ip[4];
                 ip[0] = (uint8_t)(pxe->server_ip & 0xFF);
                 ip[1] = (uint8_t)((pxe->server_ip >> 8) & 0xFF);
-                ip[3] = (uint8_t)((pxe->server_ip >> 24) & 0xFF);
                 ip[2] = (uint8_t)((pxe->server_ip >> 16) & 0xFF);
+                ip[3] = (uint8_t)((pxe->server_ip >> 24) & 0xFF);
                 px_rs232_printf("Stivale2 PXE ip addr: %d.%d.%d.%d\n",
                     ip[3], ip[2], ip[1], ip[0]
                 );
--- a/kernel/arch/i386/boot/bootparse.cpp
+++ b/kernel/arch/i386/boot/bootparse.cpp
@@ -270,11 +270,11 @@ void px_parse_stivale2(void *info)
             case STIVALE2_STRUCT_TAG_PXE_SERVER_INFO:
             {
                 auto pxe = (struct stivale2_struct_tag_pxe_server_info*)tag;
-                unsigned char ip[4];
-                ip[0] = pxe->server_ip & 0xFF;
-                ip[1] = (pxe->server_ip >> 8) & 0xFF;
-                ip[2] = (pxe->server_ip >> 16) & 0xFF;
-                ip[3] = (pxe->server_ip >> 24) & 0xFF;
+                uint8_t ip[4];
+                ip[0] = (uint8_t)(pxe->server_ip & 0xFF);
+                ip[1] = (uint8_t)((pxe->server_ip >> 8) & 0xFF);
+                ip[3] = (uint8_t)((pxe->server_ip >> 24) & 0xFF);
+                ip[2] = (uint8_t)((pxe->server_ip >> 16) & 0xFF);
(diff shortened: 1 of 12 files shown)


kernel/arch/i386/boot/bootparse.cpp
```diff
@@ -270,11 +270,11 @@ void px_parse_stivale2(void *info)
...
                 auto pxe = (struct stivale2_struct_tag_pxe_server_info*)tag;
-                unsigned char ip[4];
-                ip[0] = pxe->server_ip & 0xFF;
-                ip[1] = (pxe->server_ip >> 8) & 0xFF;
-                ip[2] = (pxe->server_ip >> 16) & 0xFF;
-                ip[3] = (pxe->server_ip >> 24) & 0xFF;
+                uint8_t ip[4];
+                ip[0] = (uint8_t)(pxe->server_ip & 0xFF);
+                ip[1] = (uint8_t)((pxe->server_ip >> 8) & 0xFF);
+                ip[3] = (uint8_t)((pxe->server_ip >> 24) & 0xFF);
```
micahswitzer: Why reverse the order of these two lines?

kernel/dev/tty/tty.cpp
```diff
@@ -46,14 +46,14 @@ void px_tty_clear(px_tty_vga_color fore, px_tty_vga_color back) {
     reset_fore = fore;
     reset_back = back;
     volatile uint16_t* where;
-    uint16_t attrib = VGA_COLOR(color_back, color_fore);
+    uint16_t attrib = (uint16_t)VGA_COLOR(color_back, color_fore);
```
micahswitzer: Can this cast be moved into the macro?

kernel/dev/tty/tty.cpp
```diff
@@ -68,7 +68,7 @@ void px_tty_reset_defaults() {
 
 void px_set_indicator(px_tty_vga_color color) {
     volatile uint16_t* where;
-    uint16_t attrib = (color << 4) | (color & 0x0F);
+    uint16_t attrib = (uint16_t)((color << 4) | (color & 0x0F));
     where = x86_bios_vga_mem + (X86_IND_Y * X86_TTY_WIDTH + X86_IND_X);
-    *where = ' ' | (attrib << 8);
+    *where = (uint16_t)(' ' | (attrib << 8));
```
micahswitzer: Can we use the macros here?

kernel/arch/i386/boot/bootparse.cpp
Kfeavel: Was accidental. Probably accidentally swapped when trying to do a keyboard macro.

Kfeavel: Resolved in fe71257.

kernel/dev/tty/tty.cpp
Kfeavel: I believe I tried that and it didn't work. The reason is that because the macro does operations (mathematical or bitwise) GCC converts it into an `int` (for some reason) so we have to cast the final result and not the individual parameters.

kernel/dev/tty/tty.cpp
Kfeavel: What do you mean?

kernel/dev/tty/tty.cpp
micahswitzer: Macros are just text replacement, so it should be lexically identical to move the cast inside the macro

kernel/dev/tty/tty.cpp
micahswitzer: `VGA_CHAR(' ', attrib)`

Kfeavel: Ah, yeah. Don't know why it wasn't like that before. Fixing.

kernel/dev/tty/tty.cpp
Kfeavel: ~~Oh, I see what you're saying. My only thing with that is it would force the macro to only be used with `uint16_t` types. Is that something we're okay with?~~

I think it makes sense so I went ahead and did it. Resolved in c1f2b9d.

kernel/dev/tty/tty.cpp
Kfeavel: Resolved in c1f2b9d

micahswitzer (approved):

PR #212, opened 2021-05-05
WIP: Resolve "Refactor Boot Process & Kernel Location"

### LINKED ISSUE #211: Refactor Boot Assembly into C
~~The kernel boot process should be refactored so that the kernel no longer resides in the higher half of memory. This will require the program loader to support dynamic relocation, but that is a compromise that we are willing to make. The boot process will be greatly simplified by not having paging code in `boot.S`. Additionally, stivale2 could be used to boot directly into parsing code before main, and a similar process can be employed for Multiboot2.~~

The new objective is to convert as much boot assembly into C as possible. All the code necessary to boot into the higher half show be located in the `.early_*` sections.

Kfeavel:
Closes #211

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -145,17 +145,18 @@ export LDFLAGS :=           \
 # * Kernel Build Targets *
 # ************************
 
+# Debug build
+debug: CPPFLAGS += -DDEBUG
+debug: CXXFLAGS += -ggdb3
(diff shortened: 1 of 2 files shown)


Kfeavel:
Closing for now. Decided to focus on using existing code to solve memory issues instead of reworking boot code.

PR #217, opened 2021-05-31
Resolve "Remove `px_` Prefix"

### LINKED ISSUE #216: Remove `px_` Prefix
This is really a two part issue, but they should be separated as we continue to reorganize the kernel. Since we're using C++, namespaces are a feature we can utilize (when appropriate. They're best avoided when interfacing with assembly or C) and thus can replace the `px_` prefix in front of all kernel functions and types.

TL;DR - The `px_` prefix is unnecessary and kind of annoying at this point.

Kfeavel:
Closes #216

### DIFF
--- a/gdb.commands
+++ b/gdb.commands
@@ -1,6 +1,6 @@
 target remote localhost:1234
 file dist/kernel
-break px_kernel_main
+break kernel_main
 set disassembly-flavor intel
 layout asm
 focus cmd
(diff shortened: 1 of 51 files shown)


Kfeavel:
Forgot to mention that I also renamed the port functions and that got included in this MR. Would be kind of annoying to separate out, and I don't think it's that big of a deal if they get merged in with this, but I understand that it's technically not related. But, when have I ever cared about that?

micahswitzer (approved):

PR #219, opened 2021-06-03
Resolve "Create Bootloader Handoff Class"

### LINKED ISSUE #218: Create Bootloader Handoff Class
Should provide a uniform way to hold bootloader information regardless of source (i.e. GRUB, Stivale, etc.)

Should not include memory map parsing at first since that is an entirely different beast.

Kfeavel:
Closes #218

### DIFF
--- a/kernel/boot/Handoff.cpp
+++ b/kernel/boot/Handoff.cpp
@@ -1,6 +1,6 @@
 /**
  * @file Handoff.cpp
- * @author Keeton Feave (keetonfeavel@cedarville.edu)
+ * @author Keeton Feavel (keetonfeavel@cedarville.edu)
  * @brief 
  * @version 0.1
  * @date 2021-06-02
@@ -105,11 +105,11 @@ Handoff::Handoff(void* handoff, uint32_t magic)
     if (magic == 0x36d76289) {
         bootProtoName = "Multiboot2";
         _bootType = Multiboot2;
-        parseMultiboot2(handoff);
+        parseMultiboot2(this, handoff);
     } else if (magic == *(uint32_t*)"stv2") {
         bootProtoName = "Stivale2";
         _bootType = Stivale2;
-        parseStivale2(handoff);
+        parseStivale2(this, handoff);
     } else {
         PANIC("Invalid bootloader information!");
     }
@@ -128,7 +128,7 @@ Handoff::~Handoff()
  * |___/\__|_|\_/\__,_|_\___/___|
  */
 
-void Handoff::parseStivale2(void* handoff)
+void Handoff::parseStivale2(Handoff* that, void* handoff)
 {
     struct stivale2_struct* fixed = (struct stivale2_struct*)handoff;
     // Walk the list of tags in the header
@@ -145,7 +145,7 @@ void Handoff::parseStivale2(void* handoff)
             {
                 auto cmdline = (struct stivale2_struct_tag_cmdline*)tag;
                 rs232_printf("Stivale2 cmdline: '%s'\n", (const char *)cmdline->cmdline);
-                _cmdline = (const char *)(cmdline->cmdline);
+                that->_cmdline = (const char *)(cmdline->cmdline);
                 break;
             }
             case STIVALE2_STRUCT_TAG_FRAMEBUFFER_ID:
@@ -158,7 +158,7 @@ void Handoff::parseStivale2(void* handoff)
                     framebuffer->framebuffer_height,
                     (framebuffer->framebuffer_bpp * 8));
                 // Initialize the framebuffer information
-                _fbInfo = FramebufferInfo(
+                that->_fbInfo = FramebufferInfo(
                     framebuffer->framebuffer_width,
                     framebuffer->framebuffer_height,
                     framebuffer->framebuffer_bpp,
@@ -199,7 +199,7 @@ struct multiboot_fixed
     uint32_t reserved;
 };
 
-void Handoff::parseMultiboot2(void* handoff)
+void Handoff::parseMultiboot2(Handoff* that, void* handoff)
 {
     auto fixed = (struct multiboot_fixed *) handoff;
     // TODO: Find a way around this when the new memory manager code is done.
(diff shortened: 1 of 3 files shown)


kernel/boot/Handoff.cpp
```diff
@@ -0,0 +1,171 @@
...
+
+FramebufferInfo::FramebufferInfo(uint16_t width, uint16_t height, uint16_t depth, uint16_t pitch, void* addr)
+    : _addr(addr)
+    , _width(width)
+    , _height(height)
+    , _depth(depth)
+    , _pitch(pitch)
+{
+    // TODO: Can we call the default constructor or will that overwrite
+    // the values passed in the initialization list?
```
micahswitzer: Not sure, but we can at least use the constructor syntax like in the default constructor.

kernel/boot/Handoff.cpp
```diff
@@ -0,0 +1,171 @@
...
+    _redMaskSize = 0;
+    _redMaskShift = 0;
+    _greenMaskSize = 0;
+    _greenMaskShift = 0;
+    _blueMaskSize = 0;
+    _blueMaskShift = 0;
+    _memoryModel = Undefined_FBMM;
+}
+
+FramebufferInfo::~FramebufferInfo()
```
micahswitzer: Do you need to specify a destructor if the type is simple and doesn't need destruction logic?

kernel/boot/Handoff.cpp
```diff
@@ -0,0 +1,171 @@
...
+{
+    // Nothing to deconstruct
+}
+
+uint16_t FramebufferInfo::getWidth() { return _width; }
+uint16_t FramebufferInfo::getHeight() { return _height; }
+uint16_t FramebufferInfo::getDepth() { return _depth; }
```
micahswitzer: Would it be a good idea to move `FramebufferInfo`'s implementation entirely into the header file so that we don't incur any performance cost by using a class?

kernel/boot/Handoff.cpp
```diff
@@ -0,0 +1,171 @@
...
+        parseStivale2(handoff);
+    }
+}
+
+Handoff::~Handoff()
+{
+    // Nothing to deconstruct
+}
+
+void Handoff::parseStivale2(void* handoff)
```
micahswitzer: Can we move this to a static builder method where it constructs and returns a `Handoff` object?

sysroot/usr/include/kernel/boot/Handoff.hpp
```diff
@@ -0,0 +1,129 @@
...
+    UEFI = 1,
+};
+
+enum HandoffBootloaderType {
+    Multiboot2 = 0,
+    Stivale2 = 1,
+};
+
+// Unused for now.
+class HandoffRSDPDescriptor {
```
micahswitzer, line 22: Does this info need to be handed off to the kernel, or can we just preserve the RSDT address?

kernel/boot/Handoff.cpp
```diff
@@ -0,0 +1,171 @@
+/**
+ * @file Handoff.cpp
+ * @author Keeton Feave (keetonfeavel@cedarville.edu)
```
micahswitzer:
```suggestion
 * @author Keeton Feavel (keetonfeavel@cedarville.edu)
```

micahswitzer (commented):
Looks pretty good. I asked a few questions that I don't have an answer to, so we'll need to discuss those.

kernel/boot/Handoff.cpp
Kfeavel: I was mostly trying to avoid duplicating that code, but yeah, it may just be best to do that.

kernel/boot/Handoff.cpp
Kfeavel: Not technically, no.

kernel/boot/Handoff.cpp
Kfeavel: Yeah, I was trying to figure out how to do that. For some reason the compiler wasn't happy when I tried, so I have to figure that out.

kernel/boot/Handoff.cpp
Kfeavel: The idea is that the constructor takes the `void* handoff` pointer and then will call either `parseStivale2` or `parseMultiboot2` and construct itself that way, so basically the same end result.

sysroot/usr/include/kernel/boot/Handoff.hpp
Kfeavel: Do we not want to parse the RSDP information? We would still keep the pointer, but I figured having a class around it to access information easier would be nice. Feel free to disagree.

kernel/boot/Handoff.cpp
Kfeavel: Resolved in 6757f8e.

kernel/boot/Handoff.cpp
Kfeavel: Resolved in 6757f8e.

kernel/boot/Handoff.cpp
Kfeavel: Resolved in 6757f8e.

kernel/boot/Handoff.cpp
micahswitzer: Oh. Calling member functions from the constructor. Nice.

sysroot/usr/include/kernel/boot/Handoff.hpp
micahswitzer: Hmm well I was just thinking it would be simpler to not have it, but I guess keep it until we know for sure what we want to do with that information.

kernel/boot/Handoff.cpp
micahswitzer: This is still wrong.

kernel/boot/Handoff.cpp
Kfeavel: Updated in 651d0e3. Let me know if you have any other thoughts on the matter.

micahswitzer (approved):

PR #230, opened 2021-06-17
Update ring buffer API to allow static buffers

Kfeavel:
Cherry-picked from #206


micahswitzer:
Any reason we can't use the template-based approach that I was suggesting earlier? I think it's an elegant way to solve this problem.

Something like the following (not sure if the template syntax is 100% correct)
```c++
template <int T>
class RingBuffer<T> {
private:
    uint8_t _data[T];
public:
    RingBuffer<T>();
}
```

Kfeavel:
> Any reason we can't use the template-based approach that I was suggesting earlier? I think it's an elegant way to solve this problem.
> 
> Something like the following (not sure if the template syntax is 100% correct)
> 
> ```c++
> template <int T>
> class RingBuffer<T> {
> private:
>     uint8_t _data[T];
> public:
>     RingBuffer<T>();
> }
> ```

Resolved in 318e5d9.
Edit: Added size to template in 4243d47

sysroot/usr/include/kernel/lib/ring_buffer.hpp
```diff
@@ -13,84 +13,129 @@
...
+    }
+    /**
+     * @brief Query whether the circular buffer is empty.
+     *
+     * @return true The buffer is empty
+     * @return false The buffer is not empty
+     */
+    bool IsEmpty()
+    {
+        return (this->length == 0);
```
micahswitzer: Remove extra parens

kernel/dev/serial/rs232.cpp
```diff
@@ -125,33 +129,11 @@ void rs232_init(uint16_t com_id) {
...
-
 char rs232_get_char() {
     mutex_lock(&mutex_rs232);
     // Grab the last byte and convert to a char
     uint8_t data = 0;
-    if (read_buffer != NULL)
-    {
-        ring_buffer_dequeue(read_buffer, &data);
-    }
+    ring.Dequeue(&data);
```
micahswitzer: Do we need to check the return code here?

kernel/dev/serial/rs232.cpp
```diff
@@ -12,17 +12,18 @@
...
+#include <lib/mutex.hpp>
+#include <lib/string.hpp>
+#include <lib/ring_buffer.hpp>
 
-static ring_buff_t* read_buffer = NULL;
-uint8_t rs_232_line_index;
-uint16_t rs_232_port_base;
-mutex_t mutex_rs232("rs232");
+static uint16_t rs_232_port_base;
+static RingBuffer<uint8_t, 1024> ring;
```
micahswitzer: Why use `uint8_t` if we're enqueuing and dequeuing into `char`s?

sysroot/usr/include/kernel/lib/ring_buffer.hpp
```diff
@@ -13,84 +13,129 @@
...
+        return status;
+    }
+    /**
+     * @brief Removes a bytes from the circular buffer and writes
+     * the value of the byte to the data pointer.
+     *
+     * @param buf Buffer to contain the data
+     * @return int Returns 0 on success and -1 on error.
+     */
+    int Dequeue(T* buf)
```
micahswitzer, line 61: Can we maybe get an option to dequeue by value?

sysroot/usr/include/kernel/lib/ring_buffer.hpp
```diff
@@ -13,84 +13,129 @@
...
+    }
+    /**
+     * @brief Query whether the circular buffer is full.
+     *
+     * @return true The buffer is full
+     * @return false The buffer is not full
+     */
+    bool IsFull()
+    {
+        return (this->length == S);
```
micahswitzer: same as above

kernel/dev/serial/rs232.cpp
Kfeavel: I feel like `uint8_t` is better for raw data, but there's no real reason. I can switch it so we can avoid a cast.

kernel/dev/serial/rs232.cpp
Kfeavel: The worst thing that happens is a null character gets returned, and I'm not sure what else we would do besides that, so I'm not sure it's really worth checking the return value. What would we do in event of failure besides return 0 / null character?

sysroot/usr/include/kernel/lib/ring_buffer.hpp
Kfeavel: Resolved in 04f2009.

kernel/dev/serial/rs232.cpp
Kfeavel: Resolved in 04f2009.

sysroot/usr/include/kernel/lib/ring_buffer.hpp
Kfeavel: Resolved in 04f2009.

sysroot/usr/include/kernel/lib/ring_buffer.hpp
Kfeavel: Resolved in 04f2009.

sysroot/usr/include/kernel/lib/ring_buffer.hpp
```diff
@@ -13,84 +13,141 @@
...
+     * @return int Buffer capacity in bytes
+     */
+    int Capacity()
+    {
+        return S;
+    }
+
+private:
+    T data[S];
     int head;
```
micahswitzer: I know this isn't part of these changes, but could these be `size_t`s instead?

Kfeavel: Resolved in f95c4f0.

micahswitzer (approved):

PR #236, opened 2021-06-21
Linked List Refactor

Kfeavel:
Converts the linked list code into a templated class like what was done with the ring buffer code. Also renames ring_buffer to RingBuffer to match naming scheme.


sysroot/usr/include/kernel/lib/LinkedList.hpp
```diff
@@ -59,7 +59,7 @@ class LinkedListNode {
      *
      * @return T Stored data
      */
-    T GetData()
+    T Data()
```
micahswitzer: Maybe make this return a `T&` for more efficient mutation (unless we want the data to immutable)

sysroot/usr/include/kernel/lib/LinkedList.hpp
```diff
@@ -0,0 +1,249 @@
...
+class LinkedList {
+public:
+    LinkedList(T val)
+    {
+        head = LinkedListNode<T>(val);
+    }
+    void PushFront(T val)
+    {
+        LinkedListNode<T> next = head;
+        LinkedListNode<T> newHead = new LinkedListNode<T>(val, next, NULL);
```
micahswitzer: Ah I think this might be a good place to use smart pointers for reference counting. I can work on an implementation if you want.

See [here](https://en.cppreference.com/w/cpp/memory/shared_ptr) for an example of what I'm thinking of.

Kfeavel: Don’t think we can do that without a standard library.

micahswitzer: Well I meant I'd work on a version for our own "standard library"

sysroot/usr/include/kernel/lib/LinkedList.hpp
Kfeavel: Resolved in 9e01cca.

sysroot/usr/include/kernel/lib/LinkedList.hpp
Kfeavel: That would be nice. Definitely would use that in a lot more places than this, but that's a different issue.

tests/test-lib/test-linkedlist.cpp
```diff
@@ -0,0 +1,100 @@
...
+    SECTION("Insertion") {
+        for (int i = 0; i < UINT8_MAX; i++) {
+            list.InsertBack(i);
+        }
+        // 0 -> 254 == 255 entries
+        REQUIRE(list.Count() == UINT8_MAX);
+    }
+    // Ensure the linked list can remove data properly
+    SECTION("Insert (Back) : Removal (Back)") {
+        // Account for the zero already in place
```
micahswitzer: I'm pretty sure these comments are out of date

tests/test-lib/test-linkedlist.cpp
```diff
@@ -0,0 +1,100 @@
...
+        for (int i = 0; i < UINT8_MAX; i++) {
+            list.InsertFront(i);
+        }
+        // 0 -> 254 == 255 entries
+        REQUIRE(list.Count() == UINT8_MAX);
+        // Remove all data
+        for (int i = 0; i < UINT8_MAX; i++) {
+            auto node = list.RemoveFront();
+            REQUIRE(node != NULL);
+            REQUIRE(node->Data() == ((UINT8_MAX - 1) - i));
```
micahswitzer: Any reason why you can't just count down like in the previous test?

sysroot/usr/include/kernel/lib/LinkedList.hpp
```diff
@@ -0,0 +1,269 @@
...
+    }
+    void InsertFront(T val)
+    {
+        if (head) {
+            InsertBefore(head, val);
+        } else {
+            head = new LinkedListNode<T>(val);
+            ++count;
+        }
+        if (!tail)
```
micahswitzer: The tail should only ever be null when the head is null

sysroot/usr/include/kernel/lib/LinkedList.hpp
```diff
@@ -0,0 +1,269 @@
...
+    }
+    void InsertBack(T val)
+    {
+        if (tail) {
+            InsertAfter(tail, val);
+        } else {
+            tail = new LinkedListNode<T>(val);
+            ++count;
+        }
+        if (!head)
```
micahswitzer: The head should only ever be null when the tail is null

micahswitzer (approved):

PR #244, opened 2021-07-03
Refactor Makefile to Share Common Variables & Fix Unit Tests

Kfeavel:
This refactor cleans up a lot of duplicated variables and make the makefile structure far more robust. I'm actually a really big fan of how this all fits together.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,19 +16,19 @@ export PROJ_NAME = panix
 export MAKEFLAGS   += --no-print-directory
 # Panix Version
 export GIT_COMMIT := "$(shell git describe --abbrev=8 --dirty --always --tags)"
-export VER_MAJOR := "0"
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -155,6 +155,7 @@ release: $(KERNEL)
 # Kernel (Linked With Libraries)
 .PHONY: $(KERNEL)
 $(KERNEL):
+	echo $(ROOT)
```
micahswitzer: This feels like it's left over from debugging

Kfeavel: Good catch. Thanks.

Kfeavel: Fixed and rebased.

micahswitzer (dismissed):

micahswitzer (approved):

PR #245, opened 2021-07-08
Add Catch2 Unit Test Library

Kfeavel:
Compilation and linking take forever since we're using the Catch2 single header version, but it's easy and works. Submoduling their repo got a bit complicated since they use CMake for everything.


Kfeavel:
CodeFactor checks are failing because of the Catch2 header. I can't figure out how to get it to ignore it, so we might need to disable a check in order to merge it.

micahswitzer (approved):

PR #249, opened 2021-07-11
Force several Bitset functions to always be inlined

micahswitzer:
This gives us a major performance boost. (also fixed a bug in `Bitset::Clear()`)


Kfeavel:
Looks good to me. 👍🏼

Kfeavel (approved):

PR #252, opened 2021-07-15
Resolve "Reorganize Header Files"

### LINKED ISSUE #215: Reorganize Header Files
Private headers should be moved to the `kernel` directory in their respective locations alongside the `.cpp` files (maybe in an include directory at the same level?), and all (eventual) public kernel headers should remain in `sysroot/usr/include/kernel` where they are now.

Kfeavel:
Closes #215


.vscode/c_cpp_properties.json
```diff
@@ -6,7 +6,7 @@
             "compilerPath": "/usr/bin/gcc",
             "includePath": [
                 "${default}",
-                "${workspaceFolder}/sysroot/usr/include/kernel/",
+                "${workspaceFolder}/kernel/",
```
micahswitzer, line 9: So can no longer use the sysroot include directory?

Kfeavel: I removed it since we don't have anything in it now. I could add it back, but I figured we were going to reorganize it / add things as necessary when the time arrives.

micahswitzer (approved):

PR #256, opened 2021-07-19
Update README.md

Kfeavel:
Removes some references to Cedarville simply because we don't want to give the impression that only Cedarville students (or students in general) can contribute. The readme should welcome everyone.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,18 +9,14 @@
 </div>
 
 ## Panix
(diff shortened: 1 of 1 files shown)


Kfeavel:
We website should also be updated in a similar manner.

micahswitzer (approved):

PR #259, opened 2021-07-21
Add a pseudo-random number function

Jmallone:
The pseudorandom number function in `kernel/lib/stdlib.c`  folder.
`rand()` and `srand()`


kernel/lib/stdlib.hpp
```diff
@@ -0,0 +1,25 @@
...
+ * @author Michel (JMallone) Gomes (michels@utfpr.edu.br)
+ * @brief Standard library
+ * @version 0.1
+ * @date 2021-07-20
+ *
+ * @copyright Copyright the Panix Contributors (c) 2021
+ *
+ */
+#pragma once
+#define MAX_RAND 32767
```
Kfeavel: Is this necessary to access outside of the source file? If not, I would move it out of the header.

Kfeavel (changes requested):
Looks good besides that one define and needing to link to the different resources (OSDev Wiki, C Standard, etc).

kernel/lib/stdlib.hpp
micahswitzer: POSIX says it should be publicly available

Kfeavel: Awesome, leave it here then. Marking as resolved.

Jmallone:
nice, i'll change this

Kfeavel:
@Jmallone I would also rename the files to `rand.hpp` and `rand.cpp` instead of `stdlib.*`. I know this isn't POSIX complient, but I don't think that's necessarily a goal right now. We want to (ideally) do things smarter than POSIX since we're able to start from the ground up.

Kfeavel:
Closed since @Jmallone couldn't update the MR to point to their new branch.

PR #260, opened 2021-07-22
Add a pseudo-random number function

Jmallone:
The pseudorandom number function in `kernel/lib/rand.cpp` folder.
`rand()` and `srand()`

### DIFF
--- a/kernel/lib/rand.cpp
+++ b/kernel/lib/rand.cpp
@@ -1,0 +1,32 @@
+/**
+ * @file rand.cpp
+ * @author Brian Kernighan, Dennis Ritchie (C Standard Authors)
+ * Michel (JMallone) Gomes (michels@utfpr.edu.br)
+ * @brief Portable implementation of rand and srand as according to the
+ * C standard implementation by K&R.
+ * @version 0.1
+ * @date 2021-07-20
+ *
+ * @copyright C Programming Language copyright Brian Kernighan, Dennis Ritchie
+ * Implementation copyright the Panix Contributors (c) 2021.
+ *
+ * References:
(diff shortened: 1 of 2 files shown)


kernel/lib/rand.cpp
```diff
@@ -0,0 +1,34 @@
...
+    _seedrand = _seedrand * 214013L + 2531011L;
+    return (unsigned int)(_seedrand >> 16) & MAX_RAND;
+}
+
+void srand(unsigned int seed)
+{
+    _seedrand = seed;
+}
+
+
```
Kfeavel: There's a bit of extra whitespace here that should be removed. Always have one empty line at the end of a file, but no more empty lines than that. Do the `fixup` technique that I talked to you about when you've made the change. This should be good practice. 😄

Kfeavel (changes requested):

Kfeavel (dismissed):
Looks good to me. Got everything rebased, and from what I can tell (on my phone), the whitespace is fixed. Letting @micahswitzer finish the review.

kernel/lib/rand.cpp
micahswitzer: Both files now have one extra line. I'm not sure which editor you're using, but VIM will add an extra line past what's shown so maybe that's what happened here.

Jmallone (commented):
Done, all empty line is over ;D

kernel/lib/rand.cpp
Jmallone: i'ts done :D

micahswitzer (approved):
Great work! Thanks for contributing!

PR #266, opened 2021-07-23
Add Framebuffer Text Printing

### LINKED ISSUE #163: Printing Functions Should be Moved to TTY
Currently the`putchar()` and `puts()` functions print to the kernel TTY directly. This behavior is not correct since the standard library code should be agnostic as to where the TTY is. Ideally, the TTY driver provides a `write()` function and the kernel manages which device the `put*()` functions write to.

### LINKED ISSUE #262: Framebuffer Font
Time to move away from using VGA text mode.

Kfeavel:
Closes #262 and #163

There's quite a bit of refactoring going on here too since the graphics subsystem is still pretty early in development.


Kfeavel:
Note to self: Figure out why first line prints correctly (0,0) but remaining lines are shifted over by a single character's width.

Kfeavel:
Note to self: Fix the panic screen not clearing the framebuffer and changing the background color.

Kfeavel:
Requesting an initial review from @micahswitzer while I am waiting on #275 to be resolved so that I can remove the old TTY width and height defines and add features like scrolling.

kernel/arch/i386/gdt.cpp
```diff
@@ -63,5 +62,4 @@ void gdt_install() {
     gdt_set_gate(4, 0, 0x000FFFFF, GDT_DATA_PL3); // User mode data segment
 
     gdt_flush((uint32_t)&gdt_ptr);
-    kprintf(DBG_OKAY "Installed the GDT.\n");
 }
```
micahswitzer: Missing trailing newline

kernel/arch/i386/idt.cpp
```diff
@@ -28,10 +28,8 @@ void idt_set_gate(int n, uint32_t handler_addr) {
 }
 
 void load_idt() {
-    kprintf(DBG_INFO "Loading the IDT...\n");
     idt_reg.base = (uint32_t) &idt;
     idt_reg.limit = IDT_ENTRIES * sizeof(idt_gate_t) - 1;
     /* Don't make the mistake of loading &idt -- always load &idt_reg */
     asm volatile("lidt (%0)" : : "r" (&idt_reg) : "memory");
-    kprintf(DBG_OKAY "Loaded the IDT.\n");
 }
```
micahswitzer: Missing trailing newline

kernel/dev/graphics/console.cpp
```diff
@@ -0,0 +1,302 @@
...
+
+// Coorinate trackers
+uint8_t cursorX = 0;
+uint8_t cursorY = 0;
+// VGA colors (defaults are white on black)
+uint32_t colorBack = VGA_Black;
+uint32_t colorFore = VGA_White;
+// Default colors set by tty_clear()
+uint32_t resetBack = VGA_Black;
+uint32_t resetFore = VGA_White;
```
micahswitzer: Can't/shouldn't these all be `static`?

kernel/dev/graphics/font.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * @brief Draw the character at the given coordinates in the given color.
+ *
+ * @param c Character to draw
+ * @param x Horizontal position
+ * @param y Vertical position
+ * @param fore Character hex color
+ */
+void Draw(char c, uint32_t x, uint32_t y, uint32_t fore);
```
micahswitzer, line 29: Can we note that these need to be protected by a mutex?

kernel/arch/i386/isr.cpp
```diff
@@ -82,9 +74,7 @@ extern "C" void irq_handler(registers_t *regs) {
 
     /* Handle the interrupt in a more modular way */
     if (interrupt_handlers[regs->int_num] != 0) {
-        set_indicator(VGA_Yellow);
         isr_t handler = interrupt_handlers[regs->int_num];
         handler(regs);
     }
-    set_indicator(VGA_Green);
 }
```
micahswitzer: Missing trailing newline

kernel/sys/trace.cpp
```diff
@@ -9,29 +9,30 @@
...
+        Console::printf("\033[%i;%iH  %s", (frame + 1), (X86_TTY_WIDTH - 16), buf);
         rs232::printf("%s", buf);
         // Check whether the address is in memory or not
         if (!page_is_present((size_t)stk->ebp))
-        {
             break;
-        }
         stk = stk->ebp;
     }
 }
```
micahswitzer: Missing trailing newline

micahswitzer (commented):
Wow, great work so far!

kernel/dev/graphics/font.hpp
Kfeavel: Should the font be locked, or should whatever is accessing the font (console, etc) be required to lock?

micahswitzer: Yeah really anything that writes to the same buffer should be locked.

Kfeavel: This will be resolved when #275 is finished since the font will be written to a specified `Screen` object. That way the font data isn't written to the master framebuffer.

Kfeavel:
@micahswitzer thoughts on getting this merged in without scrolling support / a few other things? #307 is likely going to conflict with this, so it would be nice to get this in soon, even if not technically complete.

Kfeavel:
Okay, I was going to integrate #275 into this PR, but that issue is *way* more complicated than I though it would be. If it's alright with others, it might be better to merge this now, even if it's incomplete.

If that's the case, new issues should be added for the following:
- Console needs to have `Screen` so that scrolling can be implemented
- Find a better way to clear the console (see `Screen` abstraction again)

micahswitzer (approved):

PR #269, opened 2021-07-24
Resolve "Implement GCC CXA Functions"

### LINKED ISSUE #268: Implement GCC CXA Functions
Panix should have GCC local static variable guards and the pure virtual function.

Kfeavel:


kernel/meta/gcc.cpp
```diff
@@ -0,0 +1,56 @@
...
+{
+    __atomic_clear(g, __ATOMIC_RELEASE);
+}
+
+} // !__cxxabiv1
+#endif
+
+extern "C" void __cxa_pure_virtual()
+{
+    rs232::printf("[%s] Pure virtual called!\n", __FUNCTION__);
```
micahswitzer: We *must* panic here, as this is undefined behavior. Also, `__FUNCTION__` will always be `__cxa_pure_virtual`, so I'm not sure how useful that is. The panic screen's back trace will reveal who called it.

Kfeavel: Okay, I was wondering. Originally I had a panic there, but I couldn't remember if it could happen normally. Good to know.

Kfeavel: We do need to figure out how to get symbols working in the panic backtrace.

micahswitzer: Yeah, probably should open an issue for that.

micahswitzer (approved):

PR #273, opened 2021-07-25
Binary IO Classes

micahswitzer:
These are just some classes and interfaces to make doing IO a slight bit more elegant and less error-prone.

### DIFF
--- a/kernel/lib/io.hpp
+++ b/kernel/lib/io.hpp
@@ -1,0 +1,78 @@
+#pragma once
+
+#include <stddef.h>
+#include <stdbool.h>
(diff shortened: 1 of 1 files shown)


kernel/lib/io.hpp
```diff
@@ -0,0 +1,93 @@
+#pragma once
+
+#include <stddef.h>
+#include <stdbool.h>
+#include <sys/types.h>
+
+namespace io {
+
+enum whence {
```
jamesaorson, line 9: Should this be cased as `Whence`?

PR #276, opened 2021-07-25
Resolve "Use Unix Epoch for RTC & Create time.h" 

### LINKED ISSUE #227: Use Unix Epoch for RTC & Create `time.h`
Currently there is no way to get information out of the RTC driver. It can only print it to screen. `time.h` should provide the commonly used functions as defined by the POSIX spec (found [here](https://pubs.opengroup.org/onlinepubs/007908775/xsh/time.h.html)) along with `stuct tm`. All communication with the RTC driver should be in order to get a Unix Epoch that can be used for other time conversion elsewhere.

Jmallone:
Closes #227

### DIFF
--- a/kernel/dev/rtc/rtc.cpp
+++ b/kernel/dev/rtc/rtc.cpp
@@ -1,39 +1,48 @@
 /**
  * @file rtc.cpp
  * @author Keeton Feavel (keetonfeavel@cedarville.edu)
+ * @author Michel (JMallone) Gomes (michels@utfpr.edu.br)
  * @brief
- * @version 0.3
- * @date 2019-11-18
+ * @version 0.4
+ * @date 2021-07-25
  *
- * @copyright Copyright the Panix Contributors (c) 2019
+ * @copyright Copyright the Panix Contributors (c) 2021
  *
+ * References:
+ *         https://github.com/limine-bootloader/limine/blob/trunk/stage23/lib/time.c
+ *         https://en.wikipedia.org/wiki/Julian_day
  */
-
 #include <arch/arch.hpp>
-#include <dev/tty/tty.hpp>
 #include <dev/rtc/rtc.hpp>
-#include <lib/string.hpp>
+#include <dev/tty/tty.hpp>
 #include <lib/stdio.hpp>
+#include <lib/string.hpp>
 
-void rtc_callback(registers_t *regs);
-bool rtc_get_update_in_progress();
-uint8_t rtc_get_register(uint8_t reg);
-void read_rtc();
+#define RTC_CMOS_PORT 0x70
+#define RTC_DATA_PORT 0x71
+#define RTC_CURRENT_YEAR 2021 // Needs to be updated every year!
+#define RTC_CURRENT_CENTURY 0 // Needs to be updated every century!
+
+static void rtc_callback(registers_t* regs);
+static bool rtc_get_update_in_progress();
+static uint8_t rtc_get_register(uint8_t reg);
+static void read_rtc();
 
 // Current values from RTC
 // These variables are way larger than they ever
 // should be, but the compiler doesn't like our
 // math with anything smaller, so we're going
 // to live with it since memory is "cheap"
-uint32_t rtc_second;      // Current UTC second
-uint32_t rtc_minute;      // Current UTC minute
-uint32_t rtc_hour;        // Current UTC hour
-uint32_t rtc_day;         // Current UTC day (not reliable)
-uint32_t rtc_month;       // Current UTC month
-uint32_t rtc_year;        // Current UTC year
-uint32_t rtc_century;     // Current UTC century
+uint32_t rtc_second;  // Current UTC second
+uint32_t rtc_minute;  // Current UTC minute
+uint32_t rtc_hour;    // Current UTC hour
+uint32_t rtc_day;     // Current UTC day (not reliable)
+uint32_t rtc_month;   // Current UTC month
+uint32_t rtc_year;    // Current UTC year
+uint32_t rtc_century; // Current UTC century
 
-void rtc_init() {
+void rtc_init()
+{
     kprintf(DBG_INFO "Initializing RTC...\n");
     // Initializer
     writeByte(RTC_CMOS_PORT, 0x8A);
@@ -47,17 +56,20 @@ void rtc_init() {
     register_interrupt_handler(IRQ8, rtc_callback);
 }
 
-void rtc_callback(registers_t *regs) {
+static void rtc_callback(registers_t* regs)
+{
     (void)regs;
     kprintf(DBG_INFO "RTC updated.\n");
 }
 
-bool rtc_get_update_in_progress() {
+static bool rtc_get_update_in_progress()
+{
     writeByte(RTC_CMOS_PORT, 0x0A);
     return (readByte(RTC_DATA_PORT) & 0x80);
 }
 
-uint8_t rtc_get_register(uint8_t reg) {
+static uint8_t rtc_get_register(uint8_t reg)
+{
     writeByte(RTC_CMOS_PORT, reg);
     return readByte(RTC_DATA_PORT);
 }
@@ -69,13 +81,14 @@ uint8_t rtc_get_register(uint8_t reg) {
     Note: This uses the "read registers until you get the same values twice in a row" technique
     to avoid getting dodgy/inconsistent values due to RTC updates
 */
-void read_rtc() {
+static void read_rtc()
+{
     // Previous values from RTC
     // Used as a cache to check if we should update
     uint32_t last_second = 0, last_minute = 0,
-            last_hour = 0, last_day = 0,
-            last_month = 0,  last_year = 0,
-            last_century = 0, registerB = 0;
+             last_hour = 0, last_day = 0,
+             last_month = 0, last_year = 0,
+             last_century = 0, registerB = 0;
     // Make sure an update isn't in progress
     while (rtc_get_update_in_progress());
 
@@ -87,10 +100,10 @@ void read_rtc() {
     rtc_year = rtc_get_register(0x09);
     rtc_century = rtc_get_register(RTC_CURRENT_CENTURY);
 
-    while  ((last_second != rtc_second) || (last_minute != rtc_minute) ||
-            (last_hour != rtc_hour)     || (last_day != rtc_day)       ||
-            (last_month != rtc_month)   || (last_year != rtc_year)     ||
-            (last_century != rtc_century))
+    while ((last_second != rtc_second) || (last_minute != rtc_minute) || 
+           (last_hour != rtc_hour)     || (last_day != rtc_day)       ||
+           (last_month != rtc_month)   || (last_year != rtc_year)     || 
+           (last_century != rtc_century)) 
     {
         last_second = rtc_second;
         last_minute = rtc_minute;
@@ -118,12 +131,12 @@ void read_rtc() {
 
     // Convert BCD to binary values if necessary
     if (!(registerB & 0x04)) {
-        rtc_second   = (rtc_second & 0x0F) + ((rtc_second / 16) * 10);
-        rtc_minute   = (rtc_minute & 0x0F) + ((rtc_minute / 16) * 10);
-        rtc_hour     = ((rtc_hour & 0x0F) + (((rtc_hour & 0x70) / 16) * 10)) | (rtc_hour & 0x80);
-        rtc_day      = (rtc_day & 0x0F) + ((rtc_day / 16) * 10);
-        rtc_month    = (rtc_month & 0x0F) + ((rtc_month / 16) * 10);
-        rtc_year     = (rtc_year & 0x0F) + ((rtc_year / 16) * 10);
+        rtc_second = (rtc_second & 0x0F) + ((rtc_second / 16) * 10);
+        rtc_minute = (rtc_minute & 0x0F) + ((rtc_minute / 16) * 10);
+        rtc_hour = ((rtc_hour & 0x0F) + (((rtc_hour & 0x70) / 16) * 10)) | (rtc_hour & 0x80);
+        rtc_day = (rtc_day & 0x0F) + ((rtc_day / 16) * 10);
+        rtc_month = (rtc_month & 0x0F) + ((rtc_month / 16) * 10);
+        rtc_year = (rtc_year & 0x0F) + ((rtc_year / 16) * 10);
 
         if (RTC_CURRENT_CENTURY != 0) {
             rtc_century = (rtc_century & 0x0F) + ((rtc_century / 16) * 10);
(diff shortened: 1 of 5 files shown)


kernel/dev/rtc/rtc.cpp
```diff
@@ -136,15 +139,34 @@ void read_rtc() {
...
     } else {
         rtc_year += (RTC_CURRENT_YEAR / 100) * 100;
-        if(rtc_year < RTC_CURRENT_YEAR) rtc_year += 100;
+        if (rtc_year < RTC_CURRENT_YEAR)
+            rtc_year += 100;
     }
 }
+//
+// Julian date calculation from https://en.wikipedia.org/wiki/Julian_day
+static uint64_t get_jdn(uint8_t days, uint8_t months, uint16_t years)
```
Kfeavel: I know you've only just started working on this, but I wanted to say that it's generally bad form to just copy and paste code from somewhere else. We don't want to copy the code out of Limine but instead use it as a guide and inspiration for our own project, even if that means that the result is different only in style (which would likely be the case here since this is basically just math).

For example, the Panix way of doing things would be to rename the functions as `getJulianDay` and `getUnixEpoch`. Same thing for the variables. `jdnNow` and `jdn1970` are both valid options.

Hope that makes sense. 🙂

Jmallone: oh, thanks for the tip, i'll rework in this code later.

kernel/dev/rtc/rtc.cpp
```diff
@@ -1,14 +1,17 @@
...
  * @author Keeton Feavel (keetonfeavel@cedarville.edu)
+ * @author Michel (JMallone) Gomes (michels@utfpr.edu.br)
  * @brief
- * @version 0.3
- * @date 2019-11-18
+ * @version 0.4
+ * @date 2021-07-25
  *
  * @copyright Copyright the Panix Contributors (c) 2019
+ * @copyright Copyright 2019, 2020, 2021 mintsuki and contributors.
```
Jmallone: i have a doubt about copyright mention in this case. Is right put this ?

Kfeavel: I would remove the copyright and put a reference instead once you change the code. Move the wikipedia link under the references too.

Jmallone: Okey, thanks
I'll do it later

kernel/lib/time.cpp
```diff
@@ -0,0 +1,90 @@
...
+    , _year(year)
+{
+    // Initialize nothing.
+}
+TimeDescriptor::~TimeDescriptor()
+{
+    // Nothing to deconstruct
+}
+
+void TimeDescriptor::converterEpochToDate(uint64_t epochNumber)
```
Kfeavel, line 39: Rename this parameter to just `epoch`

kernel/lib/time.cpp
```diff
@@ -0,0 +1,90 @@
...
+        getYear(),
+        getHour(),
+        getMinutes());
+}
+
+void TimeDescriptor::toDate()
+{
+    uint64_t epoch = getEpoch();
+    converterEpochToDate(epoch);
+    // TODO: for another GMT, add or subtract time zone in epoch number
```
Kfeavel, line 85: Go ahead and open an issue for this. 🙂

kernel/lib/time.hpp
```diff
@@ -0,0 +1,55 @@
...
+    ~TimeDescriptor();
+    // Getters
+    int getSeconds() { return _seconds; }
+    int getMinutes() { return _minutes; }
+    int getHour() { return _hour; }
+    int getDay() { return _day; }
+    int getMonth() { return _month; }
+    int getYear() { return _year; }
+    // Setters
+    void setSeconds(int newSeconds) { _seconds = newSeconds; }
```
Kfeavel: I know this is super nit-picky, but I would prefer the parameters be something like `sec`, `min`, `hr`, `d`, `m`, `y` instead of `newSeconds`, etc.

kernel/main.cpp
```diff
@@ -81,11 +81,12 @@ void kernel_main(void *boot_info, uint32_t magic) {
...
     interrupts_enable();
     // Print some info to show we did things right
-    rtc_print();
+    Time::TimeDescriptor time;
+    time.printDate();
     // Get the CPU vendor and model data to print
     const char *vendor = cpu_get_vendor();
     const char *model = cpu_get_model();
-    kprintf(DBG_INFO "%s %s\n", vendor, model);
+    kprintf(DBG_INFO " %s %s\n", vendor, model);
```
Kfeavel: Revert this. I think it was accidental.

kernel/lib/time.cpp
Kfeavel: Rename this parameter to `epoch`

kernel/lib/time.cpp
```diff
@@ -0,0 +1,90 @@
...
+{
+    int a, b, c, day, month, year;
+    setSeconds(epochNumber % 60);
+    epochNumber /= 60;
+    setMinutes(epochNumber % 60);
+    epochNumber /= 60;
+    setHour(epochNumber % 24);
+    epochNumber /= 24;
+
+    a = ((4 * epochNumber + 102032) / 146097 + 15);
```
Kfeavel: Could you add comments to the right side of the code that explains what each step is doing?

kernel/lib/time.cpp
```diff
@@ -0,0 +1,90 @@
...
+ * @file time.hpp
+ * @author Michel (JMallone) Gomes (michels@utfpr.edu.br)
+ * @brief Time lib code 
+ * @version 0.1
+ * @date 2021-07-27
+ *
+ * @copyright Copyright the Panix Contributors (c) 2021
+ *
+ * References:
+ *              https://github.com/sidsingh78/EPOCH-to-time-date-converter/blob/master/epoch_conv.c
```
Kfeavel: Use 4 spaces instead of 8 here. I think `vim` defaults to a tab length larger than 4, which is what we use for Panix.

kernel/dev/rtc/rtc.cpp
```diff
@@ -136,23 +139,31 @@ void read_rtc() {
...
     } else {
         rtc_year += (RTC_CURRENT_YEAR / 100) * 100;
-        if(rtc_year < RTC_CURRENT_YEAR) rtc_year += 100;
+        if (rtc_year < RTC_CURRENT_YEAR)
+            rtc_year += 100;
     }
 }
+static uint64_t getJulianDay(uint8_t days, uint8_t months, uint16_t years)
+{
+    return (1461 * (years + 4800 + (months - 14) / 12)) / 4 + (367 * (months - 2 - 12 * ((months - 14) / 12))) / 12 - (3 * ((years + 4900 + (months - 14) / 12) / 100)) / 4
```
Kfeavel: If possible, I would like to see this one line split up into multiple variables. It will get optimized by the compiler anyways, so I would rather have readability.

kernel/dev/rtc/rtc.cpp
```diff
@@ -1,14 +1,17 @@
...
- * @version 0.3
- * @date 2019-11-18
+ * @version 0.4
+ * @date 2021-07-25
  *
- * @copyright Copyright the Panix Contributors (c) 2019
+ * @copyright Copyright the Panix Contributors (c) 2021
  *
+ * References:
+ *          https://github.com/limine-bootloader/limine/blob/trunk/stage23/lib/time.c
```
Kfeavel: 4 spaces instead of 8. 🙂

Kfeavel (changes requested):
Looks great, just have some small requests. You're really starting to get a feel for the Panix code style. Good job. 🙂

kernel/lib/time.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+    int epoch_in_jnd = 2440588;                     // 1/1/1970 converted in days since year 0 on julian calendar
+    int gregorian_in_jnd = 2299161;                 // 15/10/1582 begin of gregorian calendar
+    int sum_jnd = epoch + epoch_in_jnd;             // how many days happens since the 1/1/1970
+    int count_days = sum_jnd - gregorian_in_jnd;    // how many days happens since 15/10/1582
+    count_days -= 78;                               // minus 78, for easy do math, because its stated in 1/1/1583
+    int years_tmp = count_days / 365.2425;          // one year in gregorian calendar is 365.2425 days 
+    int year = years_tmp + 1583;                    //  (my_year - 1970) + 1583 = year in gregorian alendar
+
+    float month_tmp = count_days / 365.2425;        //      That is a MOD Function with
```
Kfeavel: So fun fact, we can't actually use `float` types in the kernel.
This code seems *way* more complicated than what was used before, so what I might suggest doing is using what we had before and creating `#define`s for some of the larger numbers like `2440588`, `2299161`, etc.
I'm just very nervous about doing so much more work when the previous code seemed to be simpler...

kernel/dev/rtc/rtc.cpp
```diff
@@ -136,23 +139,41 @@ void read_rtc() {
...
+{
+    // If month is before march
+    if (months < 3) {
+        years -= 1;
+        months += 12;
     }
+    uint64_t a = years / 100;
+    uint64_t b = a / 4;
+    uint64_t c = 2 - a + b;
+    uint64_t d = 365.25 * (years + 4716);
```
Kfeavel: You can't do `double` (numbers with decimal points) math using integers. The decimal points will just get dropped.

Kfeavel (changes requested):
If I'm being honest, the previous code was much better, even if harder to understand. In your attempt to explain the code, you've ended up making it more complicated. That's not your fault though -- time is really hard. What I would recommend doing is reverting to the code you had before and breaking the big numbers into `#define`s like I recommended earlier. Once you've done that, let me know and we can go from there. 🙂

I don't want you to think you've done a bad job -- you've done fine. I just want to make sure the code is really reliable.

Kfeavel:
Saving this for later:
http://www.cs.utsa.edu/~cs1063/projects/Spring2011/Project1/jdn-explanation.html

Kfeavel:
Had @Jmallone revert his changes in 3c30b8e back to c29c6c7. The code was simpler and even though it's harder to understand, we'll add some resources and comments explaining it.

kernel/dev/rtc/rtc.cpp
```diff
@@ -69,15 +76,17 @@ uint8_t rtc_get_register(uint8_t reg) {
...
-            last_hour = 0, last_day = 0,
-            last_month = 0,  last_year = 0,
-            last_century = 0, registerB = 0;
+             last_hour = 0, last_day = 0,
+             last_month = 0, last_year = 0,
+             last_century = 0, registerB = 0;
     // Make sure an update isn't in progress
-    while (rtc_get_update_in_progress());
+    while (rtc_get_update_in_progress())
+        ;
```
Kfeavel: I know this is likely the formatter's fault, but could you revert this? (Put the semicolon on the same line as the while)

kernel/dev/rtc/rtc.cpp
```diff
@@ -87,11 +96,7 @@ void read_rtc() {
     rtc_year = rtc_get_register(0x09);
     rtc_century = rtc_get_register(RTC_CURRENT_CENTURY);
 
-    while  ((last_second != rtc_second) || (last_minute != rtc_minute) ||
-            (last_hour != rtc_hour)     || (last_day != rtc_day)       ||
-            (last_month != rtc_month)   || (last_year != rtc_year)     ||
-            (last_century != rtc_century))
-    {
+    while ((last_second != rtc_second) || (last_minute != rtc_minute) || (last_hour != rtc_hour) || (last_day != rtc_day) || (last_month != rtc_month) || (last_year != rtc_year) || (last_century != rtc_century)) {
```
Kfeavel: Same with this line, put it back how it was before. 🙂

kernel/dev/rtc/rtc.cpp
```diff
@@ -101,7 +106,8 @@ void read_rtc() {
         last_century = rtc_century;
 
         // Make sure an update isn't in progress
-        while (rtc_get_update_in_progress());
+        while (rtc_get_update_in_progress())
```
Kfeavel: See previous comment.

kernel/dev/rtc/rtc.cpp
```diff
@@ -136,23 +142,52 @@ void read_rtc() {
...
-void rtc_print() {
+/*
+* JDN is time since noon Universal Time on January 1, 4713 BCE (on the Julian calendar)
+* Developed by Fliegel and van Flandern (1968), published in "Compact Computer Algorithms"
+* 
+*           (153 * m) + 2                y     y     y
+* JDN = d + ------------- + (365 * y) + --- - --- + --- - 32045
+*                  5                     4    100   400
+*           <  Part #1  >  < Part #2 >  <   Part #3   > < Part #4 >
+
```
Kfeavel: Missing a `*` here and at the beginning (`/**` instead of `/*`). Otherwise looks good!

kernel/dev/rtc/rtc.hpp
```diff
@@ -1,28 +1,25 @@
...
-#define RTC_CURRENT_YEAR    2019    // Needs to be updated every year!
-#define RTC_CURRENT_CENTURY 0       // Needs to be updated every century!
-
-extern uint32_t rtc_second; // Current UTC second
-extern uint32_t rtc_minute; // Current UTC minute
-extern uint32_t rtc_hour;   // Current UTC hour
-extern uint32_t rtc_day;    // Current UTC day (not reliable)
-extern uint32_t rtc_month;  // Current UTC month
-extern uint32_t rtc_year;  // Current UTC year
+#define RTC_CMOS_PORT 0x70
```
Kfeavel: Can you move these `#defines` into `rtc.cpp`? I don't think they're used anywhere else in the code.

Kfeavel (changes requested):
Looks good! Just a few small formatting changes. Some of them aren't your fault -- we still need to tweak the `clang-format` a little. 🙂

kernel/dev/rtc/rtc.hpp
Jmallone: Sure :)

kernel/dev/rtc/rtc.cpp
Jmallone: I think I should format it part by part rather than all code.

kernel/dev/rtc/rtc.cpp
Kfeavel: @Jmallone Still have to do this one. Make sure the others are lined up too. 🙂 

This is the last issue! Then it'll be good to merge!

kernel/dev/rtc/rtc.cpp
```diff
@@ -47,17 +56,20 @@ void rtc_init() {
     register_interrupt_handler(IRQ8, rtc_callback);
 }
 
-void rtc_callback(registers_t *regs) {
+void rtc_callback(registers_t* regs)
```
Kfeavel: Forgot to mention this earlier -- any function not defined in the header should be static. Besides this and the tabs, we're good!

Kfeavel (changes requested):

kernel/dev/rtc/rtc.cpp
```diff
@@ -87,10 +100,10 @@ void read_rtc() {
     rtc_year = rtc_get_register(0x09);
     rtc_century = rtc_get_register(RTC_CURRENT_CENTURY);
 
-    while  ((last_second != rtc_second) || (last_minute != rtc_minute) ||
-            (last_hour != rtc_hour)     || (last_day != rtc_day)       ||
-            (last_month != rtc_month)   || (last_year != rtc_year)     ||
-            (last_century != rtc_century))
+    while ((last_second != rtc_second) || (last_minute != rtc_minute) ||
```
Kfeavel, line 103: I'm going to go ahead and approve this since this doesn't really need to be fixed, but trailing whitespace should be trimmed. I'm not sure if there's a vim plugin for that (I'm almost certain there is though), but I would recommend installing one. That way extra whitespace is trimmed away when you save the file.

Kfeavel (approved):

PR #280, opened 2021-07-31
Resolve "Add Support For PanixBuild 3.2.0"

Kfeavel:
Close #279 

Planning on attempting to resolve #277

PR #283, opened 2021-08-05
#280 Hot-patch (Additional Fixes)

Kfeavel:
Unfortunately there were even more issues I had missed. Apparently certain sequences of commands don't like to play nice. Adding the `MODE` variable introduced a ton of hidden issues... now I see why people use SCons.

This simplifies the `run` target to no longer depend upon a boot image (file must still exist, but it isn't managed by Make). This should eventually be reverted once I figure out how to get X11 forwarding working so that Qemu can be run through Scuba, but that's being problematic on Linux, and the Scuba MR for Docker arguments in the `.scuba.yml` file is still in review.

This also fixes the `gdb.commands` file.

### DIFF
--- a/gdb.commands
+++ b/gdb.commands
@@ -1,5 +1,5 @@
 target remote localhost:1234
-file dist/kernel
+file dist/debug/kernel
 break kernel_main
 set disassembly-flavor intel
 layout asm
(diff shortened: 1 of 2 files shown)


gdb.commands
```diff
@@ -1,5 +1,5 @@
 target remote localhost:1234
-file dist/kernel
+file dist/debug/kernel
```
micahswitzer, line 2: Some or all of these can be moved to the command line, that would probably make everything easier now that I think about it

Kfeavel: What do you mean? I think it's fine to have it default to `dist/debug/kernel` since that's the only one with debug symbols.

PR #299, opened 2021-08-14
Resolve "Kernel Argument Parser"

### LINKED ISSUE #290: Kernel Argument Parser
Stivale2 and Multiboot2 both allow command line arguments to be passed to the kernel. We should devise a system where callbacks can be registered to an argument parser. This will help areas like controlling debug output levels as well as enabling or disabling features.

Kfeavel:
Closes #290

### DIFF
--- a/kernel/mem/paging.cpp
+++ b/kernel/mem/paging.cpp
@@ -57,7 +57,9 @@ static inline void set_page_dir(uint32_t page_directory);
 static inline void paging_enable();
 static inline void paging_disable();
 static void paging_args_cb(const char* arg);
-static CONSTRUCTOR void paging_args_init();
+
+// Kernel cmdline argument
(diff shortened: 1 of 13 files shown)


kernel/mem/paging.cpp
```diff
@@ -50,6 +56,11 @@ static inline void map_kernel_page_table(uint32_t pd_idx, page_table_t *table);
 static inline void set_page_dir(uint32_t page_directory);
 static inline void paging_enable();
 static inline void paging_disable();
+static void paging_args_cb(const char* arg);
+static CONSTRUCTOR void paging_args_init();
```
micahswitzer: This function isn't defined anywhere, remove the declaration

micahswitzer (approved):

kernel/mem/paging.cpp
Kfeavel: Thanks, missed that in the refactor.

micahswitzer (approved):

Issue #313, opened 2021-09-07
Refactor Memory Management


Kfeavel:
- [x] code style cleanup & general refactors (#314)
- [ ] allocating page tables in a platform independent way
- [ ] managing platform-specific pages and walking paging structures
- [ ] high-level virtual page allocator
- [ ] task separate page tables
 
(@micahswitzer correct me on these if I'm wrong / missing any)

PR #318, opened 2021-09-14
(Partially) Resolve "Refactor Memory Management (Allocating Page Tables in a Platform Independent Way)"

Kfeavel:
Addresses #313 (Task 2: allocating page tables in a platform independent way)


Kfeavel:
Due to the nature of reworking the architecture for paging, there's a lot of tangentially related code here. I know it's technically not paging related, but it's yak shaving. ¯\\_(ツ)_/¯

Kfeavel:
Going to ask @micahswitzer to review this so I can merge it in soon even though it isn't 100% complete. That way these changes get into `dev` and don't conflict as more work is done.

TL;DR I don't want this PR to get too big.

.clang-format
```diff
@@ -11,3 +11,4 @@ BreakBeforeBraces: Custom
 BraceWrapping:
     AfterFunction: true
 NamespaceIndentation: None
+IndentCaseLabels: true
```
micahswitzer, line 14: Interesting, I actually prefer them un-indented. But I can accept this

.vscode/tasks.json
```diff
@@ -1,18 +1,123 @@
...
+            "args": [
+                "make",
+                "unit-test"
+            ],
+            "group": "test"
+        },
+        {
+            "type": "shell",
+            "label": "Run (Unit Tests)",
+            "command": "scuba",
```
micahswitzer, line 56: Why does this test need `scuba` to run?

.vscode/tasks.json
```diff
@@ -1,18 +1,123 @@
...
-                "cwd": "${workspaceRoot}"
-            }
+                "env": {
+                    "MODE": "debug"
+                }
+            },
+            "problemMatcher": []
+        },
+        {
+            "type": "shell",
```
micahswitzer, line 84: Isn't there a way to make VSCode automatically connect to it?

kernel/arch/i386/Arch.i386.cpp
```diff
@@ -33,50 +33,64 @@ const char* exception_descriptions[32][16] = {
...
+    cr0.pagingEnable = 0;
     Registers::writeCR0(cr0);
 }
 
-void pageInvalidate(void* pageAddr)
+void pageInvalidate(void* addr)
 {
-   asm volatile("invlpg (%0)" ::"r" (pageAddr) : "memory");
+   asm volatile("invlpg (%0)" ::"r" (addr) : "memory");
+}
```
micahswitzer: Probably just something we can address later. But I feel like this should be `inline` in the header.

kernel/arch/i386/Arch.i386.cpp
```diff
@@ -33,50 +33,64 @@ const char* exception_descriptions[32][16] = {
...
+
+uintptr_t pageAlign(size_t addr)
+{
+    return addr & PAGE_ALIGN;
+}
+
+bool pageIsAligned(size_t addr)
+{
+    return ((addr % ARCH_PAGE_SIZE) == 0);
+}
```
micahswitzer: Same thing with moving to the header and `inline`-ing.

kernel/arch/i386/regs.hpp
```diff
@@ -29,7 +40,12 @@ struct CR0
     uint32_t reservedC          : 10;   // Reserved
     uint32_t nonWriteThrough    : 1;    // Disable write through caching?
     uint32_t cacheDisable       : 1;    // Cache disabled?
-    uint32_t paging             : 1;    // Enable paging?
+    uint32_t pagingEnable       : 1;    // Enable paging
+};
+
+struct CR2
+{
+    uint32_t pageFaultAddr      : 32;   // Address where page fault occured
```
micahswitzer, line 48: Why is this written as a bitfield?

kernel/boot/Handoff.hpp
```diff
@@ -94,20 +93,22 @@ class Handoff {
...
     // Getters
-    const char* getCmdLine()                    { return _cmdline; }
-    const void* getHandle()                     { return _handle; }
-    Graphics::Framebuffer getFramebufferInfo()  { return _framebuffer; }
-    HandoffBootloaderType getBootType()         { return _bootType; }
+    const char* CmdLine()                       { return m_cmdline; }
+    const void* Handle()                        { return m_handle; }
+    Graphics::Framebuffer* FramebufferInfo()    { return &m_framebuffer; }
+    HandoffBootloaderType* BootType()           { return &m_bootType; }
+    Memory::MemoryMap* MemoryMap()              { return &m_memoryMap; }
```
micahswitzer, line 100: Any reason we can't make these C++ references? Maybe this is a larger discussion we need to have.

kernel/dev/graphics/console.cpp
```diff
@@ -216,17 +216,19 @@ static int putchar(unsigned c, void** ptr)
     }
 
     // Move to the next line
-    if (cursorX >= X86_TTY_WIDTH) {
+    // TODO: Get width of "screen" (replace 80 with width) (#275)
+    if (cursorX >= 80) {
```
micahswitzer, line 220: Can we use the defines for now?

kernel/lib/LinkedList.hpp
```diff
@@ -257,6 +257,16 @@ class LinkedList {
     {
         return count;
     }
+    /**
+     * @brief Check if the linked list is empty
+     *
+     * @return true The list is empty
+     * @return false The list is not empty
+     */
+    bool Empty()
```
micahswitzer: I prefer the `Is` prefix for boolean-returning functions, but I'll let you decided what you want.

kernel/mem/heap.cpp
```diff
@@ -20,29 +20,22 @@ extern "C" {
 
 int liballoc_lock()
 {
-    int old_errno = errno;
-    int result = lock.Lock();
-    errno = old_errno;
-    return result;
+    return lock.Lock();
```
micahswitzer, line 23: Why do we not need to preserve the errno anymore?

kernel/mem/paging.cpp
```diff
@@ -92,35 +98,33 @@ static inline void map_kernel_page_table(uint32_t pd_idx, struct page_table* tab
...
+static void initDirectory()
 {
     // For every page in kernel memory
     for (int i = 0; i < PAGE_ENTRIES - 1; i++) {
-        map_kernel_page_table(i, &page_tables[i]);
+        mapKernelPageTable(i, &pageTables[i]);
         // clear out the page tables
         for (int j = 0; j < PAGE_ENTRIES; j++) {
-            page_tables[i].pages[j] = (struct page_table_entry) { /* ZERO */ };
+            pageTables[i].pages[j] = (struct TableEntry) { /* ZERO */ };
```
micahswitzer: Probably not something that needs to be addressed now, but it might be better to just `memset(0)` this region.

kernel/sys/trace.cpp
```diff
@@ -19,19 +19,20 @@
 void stack_trace(size_t max)
 {
     // Define our stack
-    struct stackframe* stk;
+    struct Arch::stackframe* stk;
     asm volatile("movl %%ebp, %0"
                  : "=r"(stk)::);
-    Console::printf("\033[0;%iH \033[31mStack trace:\033[0m\n", (X86_TTY_WIDTH - 16));
+    // TODO: Get width of "screen" (replace 80 with width) (#275)
+    Console::printf("\033[0;%iH \033[31mStack trace:\033[0m\n", (80 - 16));
```
micahswitzer, line 26: Again, any reason why we can't use the width `#define` in the meantime?

.vscode/tasks.json
Kfeavel: I figured it would be better to run the tests in a standard environment and not depend on possible differences in OS implementations.

.vscode/tasks.json
Kfeavel: Most likely, yes. Haven't been able to test since GDB is broken on macOS.

kernel/arch/i386/Arch.i386.cpp
Kfeavel: How often is this likely to be called?

kernel/arch/i386/regs.hpp
Kfeavel: To be consistent with the rest. I know it's a 1:1, but I figured it would be better to be consistent.

kernel/boot/Handoff.hpp
Kfeavel: They likely can be, I just didn't know much about references until pretty recently.

kernel/dev/graphics/console.cpp
Kfeavel: Ideally they'll be replaced soon™ by @Jmallone's work on the framebuffer screen abstraction. I can revert it if you'd prefer.

kernel/mem/heap.cpp
Kfeavel: `errno` is now an enum so we can't save the result to `errno` anymore. I figured at some point soon we could just update our `liballoc` implementation to use `KernelError`

kernel/mem/paging.cpp
Kfeavel: I would agree.

kernel/dev/graphics/console.cpp
Kfeavel: Plus, I'm trying to remove architecture specific defines as much as possible for now in favor of a unified architecture API / set of `#define`s.

kernel/arch/i386/Arch.i386.cpp
micahswitzer: Whenever a page is freed/unmapped

Kfeavel: No clue why but making these `inline` is causing a triple fault.

Kfeavel: [changes.txt](https://github.com/XyrisOS/xyris/files/7414308/changes.txt)

Kfeavel: Turns out it was my ~~last commit~~ tasking preparation commit. Thought I tested that...

Kfeavel: Updated and pushed.

kernel/boot/Handoff.hpp
Kfeavel: I think this needs to be a larger discussion. I'm not sure of all of the pros and cons, and it seems like it will be more than a simple "replace Class* with Class&" so it should probably be a different PR. Mind opening up a discussion?

micahswitzer (approved):

PR #322, opened 2021-10-28
Resolve "Refactor Memory Management" (Parse Bootloader Memory Map)

### LINKED ISSUE #328: Add `%ll` and `%z` support to `printf`
We use `size_t` quite a bit and having 64-bit value parsing is important.

### LINKED ISSUE #329: Fix Panic Message Formatting String
https://github.com/XyrisOS/xyris/blob/380b7d130b3a7a4239654c3e0a666f4ce82c8ed5/kernel/arch/i386/panic.cpp#L54

Failed to refactor properly a while ago. This is why the panic screen has been printing garbage lately.

Kfeavel:
Addresses part of #313 

Closes #328 and #329 (because I shaved a few yaks in the process of working on this branch)


micahswitzer:
I'm worried that removing `size_t` in many places is going to hurt portability to 64-bit systems whenever we decide to do that

Kfeavel:
> I'm worried that removing `size_t` in many places is going to hurt portability to 64-bit systems whenever we decide to do that

That's why I added the "%z" flag to printf. I'm going to put them back now that I realized it was relatively easy to add 64 bit support to printf. The formatting is a little bit different (%L vs %ll) but it does work.

kernel/arch/i386/boot/boothelp.cpp
```diff
@@ -38,23 +38,23 @@ stivale2_mmap_helper(void* baseaddr)
...
-                    if (((uint32_t)memmap->memmap[i].base) == (uint32_t)baseaddr) {
-                        return (uint32_t)memmap->memmap[i].length;
-                    }
-                    break;
+            case STIVALE2_STRUCT_TAG_MEMMAP_ID: {
+                auto memmap = (struct stivale2_struct_tag_memmap*)tag;
+                for (size_t i = 0; i < memmap->entries; i++) {
+                    switch (memmap->memmap[i].type) {
+                        case STIVALE2_MMAP_BOOTLOADER_RECLAIMABLE:
+                            if (((uint32_t)memmap->memmap[i].base) == (uint32_t)baseaddr) {
```
Kfeavel: Note to self: Convert these `uint32_t` values to `size_t` for compatibility.

Kfeavel:
Merging since nothing is panicking and memory management will still be actively worked on anyways. Shouldn't be any stability issues, but it isn't going into `stable` so we'll still be able to correct anything.

Issue #319, opened 2021-09-19
Replace GRUB with Limine for ISO


Kfeavel:
https://github.com/limine-bootloader/limine#biosuefi-hybrid-iso-creation

Kfeavel:
If I'm being honest, I don't think we need an ISO image anymore. The current image we build works just fine, and no one boots from optical media anymore.

PR #325, opened 2021-10-29
Switch to SCons

### LINKED ISSUE #324: Switch to SCons
It's time. We're done maintaining a massive `make`-based build system. Now is the time to use the glorious SCons build system.

micahswitzer:
It's time.

Closes #324


SConstruct
```diff
@@ -0,0 +1,55 @@
...
+    VERSION_NAME='Pheonix',
+    CFLAGS=[
+        '-nostdlib', '-nodefaultlibs', '-ffreestanding', '-fstack-protector-all',
+        '-fno-builtin', '-fno-omit-frame-pointer'
+    ],
+    CXXFLAGS=[
+        '-fpermissive', '-fno-rtti', '-fno-exceptions', '-fno-use-cxa-atexit',
+    ],
+    CPPDEFINES={
+        'REPO_URL': '\\"$REPO_URL\\"',
```
Kfeavel, line 101: We might want to use `shlex.quote()` for these. Just a suggestion, take it or leave it.

micahswitzer: Ooh yeah, good idea. I already had some interesting ideas for passing through these build variables as defines.

Kfeavel:
When you add the `run` task back in like we have in the `Makefile`, can you add the following to the Qemu options?

`-monitor telnet:127.0.0.1:1234,server,nowait`

SConstruct
Kfeavel: I tried using `shlex` and wasn't able to get it to work. I think it might have something to do with the fact that we're trying to access variables that aren't yet "commited" to the environment.

Kfeavel:
Remaining tasks:

- [x] Install Limine to ~~echFS~~ ext2 image (create Limine SCons tool)
- [ ] ~~Create Limine ISO~~ This can wait until #319
- [x] Build & run unit tests (with native toolchain)
- [x] Get git commit short hash
- [x] Move i686 toolchain to SCons tool
- [x] Update `Kernel` SConscript to handle `Arch` directory
- [x] `qemu` run phony targets
- [x] Documentation generation (Doxygen)

Kfeavel:
Is there a way to make targets that the user can invoke like in the Makefile? We need a way to run the kernel in Qemu, so if we can't find a way to do that with SCons, we'll likely need to keep the Makefile.

SConstruct
micahswitzer: Hmm yeah, I guess I'll have to play with it a bit. I think it might be possible with a little bit of creativity. Thinking about it a little more I don't think it's really necessary since there shouldn't be any adversarial input to the build system, but I'm not opposed to using it if we can get it to work.

micahswitzer:
> Is there a way to make targets that the user can invoke like in the Makefile? We need a way to run the kernel in Qemu, so if we can't find a way to do that with SCons, we'll likely need to keep the Makefile.

Yeah that should be possible with a `Command`

Kfeavel:
Before merging, the `liballoc` repo should have the `scons` branch merged in and this branch update the submodule reference to `xyris` (with the newly merged `SConscript`)

Scones/glob.py
```diff
@@ -0,0 +1,34 @@
+def glob(env, root: str, extensions: list):
+    """
+    Finds all files in a root directory matching the provided file extensions
+
+    Returns a flattened list of File nodes
+    """
+    sources = []
+    for (dirpath, dirs, files) in os.walk(root):
+        for extension in extensions:
+            sources.append(env.Glob(dirpath + '/*' + extension))
```
micahswitzer: This should have slightly better performance
```suggestion
            sources.append(env.Glob(f'{dirpath}/*{extension}'))
```

Scones/glob.py
```diff
@@ -0,0 +1,34 @@
+def glob(env, root: str, extensions: list):
```
micahswitzer: This is *very* close to the builtin `Glob` function. Perhaps we could call ours `RecursiveGlob`?

Kfeavel: Addressed in a9bfd261b36f4f272ff1412e79cecbe6b2258d22

Scones/glob.py
Kfeavel: Addressed in a9bfd261b36f4f272ff1412e79cecbe6b2258d22

Kfeavel (approved):

PR #338, opened 2021-11-29
Resolve "Use C++ Attribute Specifiers When Possible"

### LINKED ISSUE #335: Use C++ Attribute Specifiers When Possible
This is the correct, C++, way of doing things instead of using GCC specific attributes.
When C linkage is required, the `__attribute__` formatting will still be required since the C++ attribute specifiers won't work.

Kfeavel:
Resolves #335


Kfeavel:
This needs tested before merging. I haven't gotten my MacBook Pro to the point where it can compile a bootable image without Docker.

micahswitzer:
Why remove the `#define`s when you could just replace the `__attribute__` syntax with the `[[attribute]]` syntax? I think what you have is worse than what it's replacing.

Kfeavel:
I prefer the C++ style `[[attribute]]` over `__attribute__` primarily because it's a lot shorter. Plus, we can't just replace the macros with `[[attribute]]` since that doesn't work for functions with C linkage.

PR #342, opened 2021-12-04
Resolve "Refactor Boot Assembly into C"

### LINKED ISSUE #211: Refactor Boot Assembly into C

Kfeavel:
Addresses #211 and #346


Kfeavel:
This would be an opportune time to also address #319. Removing GRUB in favor of only Limine will help reduce to boot complexity.

Kfeavel:
> This would be an opportune time to also address #319. Removing GRUB in favor of only Limine will help reduce to boot complexity.

GRUB support removed in commit 90c0054592f0cdb0021d15b221eea81c5259dda8.

Kernel/Arch/i686/Bootloader/stivale2.c
```diff
@@ -0,0 +1,302 @@
...
+ *
+ */
+__attribute__((section(".early_text")))
+static void stage1MapLowMemory(void)
+{
+    // WARNING: code assumes that the kernel won't be greater than 3MB
+    struct DirectoryEntry* lowMem = &pageDirectory.entries[0];
+    lowMem->present = 1;
+    lowMem->readWrite = 1;
+    lowMem->tableAddr = (uint32_t)&lowMemoryPageTable;
```
micahswitzer: This needs to be right-shifted 12 bits

Kernel/Arch/i686/Bootloader/stivale2.c
```diff
@@ -0,0 +1,302 @@
...
+
+    // Map in the entirety of low-memory and stage1
+    size_t lowMemoryIdx = 0;
+    for (uintptr_t addr = ARCH_PAGE_SIZE; addr < EARLY_KERNEL_END; addr += ARCH_PAGE_SIZE) {
+        struct TableEntry* lowMemoryTableEntry = &lowMemoryPageTable.pages[lowMemoryIdx++];
+        lowMemoryTableEntry->present = 1;
+        lowMemoryTableEntry->readWrite = 1;
+        lowMemoryTableEntry->frame = addr - _KERNEL_BASE;
```
micahswitzer: I believe this address also need to be shifted 12 bits to the right.

Kernel/Arch/i686/Bootloader/stivale2.c
```diff
@@ -0,0 +1,302 @@
...
+
+    // First kernel page table entry
+    struct DirectoryEntry* kernelDirectoryEntry1 = &pageDirectory.entries[kernelDirectoryEntryIdx];
+    kernelDirectoryEntry1->present = 1;
+    kernelDirectoryEntry1->readWrite = 1;
+    kernelDirectoryEntry1->tableAddr = (uint32_t)&kernelPageTable;
+
+    // Second kernel page table entry
+    kernelDirectoryEntryIdx++;
+    struct DirectoryEntry* pagesDirectoryEntry = &pageDirectory.entries[kernelDirectoryEntryIdx];
```
micahswitzer:
```suggestion
    struct DirectoryEntry* pagesDirectoryEntry = kernelDirectoryEntry + 1;
```

Kernel/Arch/i686/Bootloader/stivale2.c
```diff
@@ -0,0 +1,302 @@
...
+
+/**
+ * @brief Prepare a higher-half stack for kernel usage.
+ *
+ */
+__attribute__((section(".early_text")))
+static void stage1JumpToStage2(void)
+{
+    // zero the kernel BSS (higher half stack)
+    for (size_t i = 0; i < _BSS_SIZE; i++) {
```
micahswitzer: You need to take the address of these symbols, otherwise you're just trying to access the value at that address.

Kernel/Arch/i686/Bootloader/stivale2.c
```diff
@@ -167,20 +168,20 @@ static void stage1MapBootloader(void)
...
     struct stivale2_tag *tags = (struct stivale2_tag *)(uintptr_t)stivale2Info->tags;
     uint32_t stivale2InfoLength = stivale2_mmap_helper(tags);
     uint32_t stivale2InfoEnd = stivale2InfoAddr + stivale2InfoLength;
 
     // Map in the entire bootloader information linked list
-    size_t bootMemoryIdx = 0;
     for (uintptr_t addr = stivale2InfoAddr; addr < stivale2InfoEnd; addr += ARCH_PAGE_SIZE) {
-        struct TableEntry* bootTableEntry = &bootPageTable.pages[bootMemoryIdx++];
+        size_t bootMemoryIdx = addr >> PAGE_TABLE_ENTRY_SHIFT;
+        struct TableEntry* bootTableEntry = &bootPageTable.pages[bootMemoryIdx & 0xfff];
```
Kfeavel: I feel like maybe we should explain why we're doing `& 0xFFF` here (because it isn't immediately obvious to me)

Kernel/Arch/i686/Bootloader/stivale2.c
```diff
@@ -193,26 +194,26 @@ static void stage1MapBootloader(void)
...
 
-    size_t kernelMemoryIdx = 0;
     for (uintptr_t addr = KERNEL_START; addr < KERNEL_END; addr += ARCH_PAGE_SIZE) {
-        struct TableEntry* kernelMemoryTableEntry = &lowMemoryPageTable.pages[kernelMemoryIdx++];
+        size_t kernelMemoryIdx = addr >> PAGE_TABLE_ENTRY_SHIFT;
+        struct TableEntry* kernelMemoryTableEntry = &kernelPageTable.pages[kernelMemoryIdx & 0xfff];
         kernelMemoryTableEntry->present = 1;
         kernelMemoryTableEntry->readWrite = 1;
-        kernelMemoryTableEntry->frame = addr;
+        kernelMemoryTableEntry->frame = kernelMemoryIdx - 0xc0000;
```
Kfeavel: Should probably make a `#define` for the `0xC000`. Is there some way we could get that from the linker script?

Kfeavel: Noting that `(KERNEL_BASE >> PAGE_TABLE_ENTRY_SHIFT)` is the correct replacement for `0xC0000`.

Kernel/Arch/Arch.hpp
```diff
@@ -24,6 +24,7 @@
  * @param info Bootloader information structure
  * @param magic Bootloader magic
  */
+// TODO: Remove extern "C" when boot.s is gone
```
micahswitzer: boot.s is gone now, but I believe this is still necessary

Kernel/Arch/i686/Bootloader/stivale2.c
```diff
@@ -0,0 +1,314 @@
...
+/**
+ * @brief identity map from 0x00000000 -> LOWMEM_END
+ *
+ */
+__attribute__((section(".early_text")))
+static void stage1MapLowMemory(void)
+{
+    // No need to get an index here since we're starting at 0x00000000,
+    // which corresponds to the first entry in the page directory.
+    // WARNING: code assumes that the kernel won't be greater than 3MB
```
micahswitzer: I'm fairly certain this is 7MB as that's what the linker script says

Kernel/Arch/i686/Bootloader/stivale2.c
```diff
@@ -0,0 +1,314 @@
...
+    // stack pointer we don't lose the one thing we need to pass into
+    // stage2 / the higher half stack
+    stivale2Info = info;
+
+    stage1MapLowMemory();
+    stage1MapHighMemory();
+    stage1MapBootloader();
+    stage1EnablePaging();
+    stage1JumpToStage2();
+    EarlyPanic("Error: Execution returned to stage1!");
```
micahswitzer: I'm curious as to why the compiler doesn't complain that this is unreachable code. Either way it is impossible for execution to ever get here.

Kernel/Arch/i686/Memory.i686.hpp
```diff
@@ -167,21 +183,38 @@ class Address
...
+ * @param pageDirPtr Address of page directory structure to be used
+ */
+__attribute__((always_inline))
+static inline void setPageDirectory(uintptr_t pageDirPtr)
 {
-    asm volatile("mov %0, %%cr3" ::"b"(page_dir));
+    asm volatile(
+        "mov %0, %%cr3"
+        :
+        : "b" (pageDirPtr)
```
micahswitzer, line 219: I know it wasn't here before, but I feel like this should have the `memory` clobber since it uh kinda changes the memory layout.

micahswitzer: Actually this is probably something that should be resolved separately

Kernel/Stacktrace.cpp
```diff
@@ -34,8 +34,8 @@ void printTrace(size_t max)
         ksprintf(buf, "0x%08X\n", stk->eip);
         Console::printf("%s", buf);
         RS232::printf("%s", buf);
-        // Check whether the address is in memory or not
-        if (!Memory::isPresent((uintptr_t)stk->ebp)) {
+        // Check for a NULL stack frame to indicate the end
+        if ((uintptr_t)stk->ebp == 0x00000000) {
```
micahswitzer, line 38: Should probably fix the `isPresent` method at some point to avoid triple faulting on a bad stack

Makefile
```diff
@@ -103,7 +102,7 @@ export CFLAGS :=            \
 	-nostdlib               \
 	-nodefaultlibs          \
 	-ffreestanding          \
-	-fstack-protector-all   \
+	-fstack-protector       \
```
micahswitzer, line 105: 🎉

Kernel/Arch/Arch.hpp
micahswitzer: Wait no, it's called from stage 2 which is C++

micahswitzer (approved):

PR #354, opened 2022-01-03
Resolve "Cleanup CPU Source (GDT, ISR, IRQ, IDT, etc.)"

### LINKED ISSUE #353: Cleanup CPU Source (GDT, ISR, IRQ, IDT, etc.)
A lot of the source is severely outdated and in need of cleanup. Given that some of it (most of it?) was written following tutorials as well, it isn't the highest quality. There are probably large chunks of code that could be removed, and if not, should be commented and explained.

Kfeavel:
Closes #353


micahswitzer (commented):
Approved.

PR #368, opened 2022-01-25
Resolve "Replace liballoc with Custom Heap"

### LINKED ISSUE #367: Replace liballoc with Custom Heap
My plan is basically to just reimplement liballoc in C++. This should be fine since liballoc is in the public domain. I'll leave some note in the header about it being a reimplementation.

Kfeavel:
Closes #367


Kernel/Library/string.cpp
```diff
@@ -138,11 +138,27 @@ void* memmove(void* destptr, const void* srcptr, size_t size)
...
+void* memcpy(void* dst, const void* src, size_t size)
 {
-    unsigned char* dst = (unsigned char*)dstptr;
-    const unsigned char* src = (const unsigned char*)srcptr;
-    for (size_t i = 0; i < size; i++)
-        dst[i] = src[i];
-    return dstptr;
+    uint8_t *bDst;
+    uint8_t *bSrc;
+    uint32_t *lDst = (uint32_t*)dst;
```
micahswitzer: Why not use `size_t`'s here?

Kernel/Memory/Heap.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+#define ALIGNMENT   16 // Memory byte alignment
+#define ALIGN_TYPE  uint8_t
+#define ALIGN_INFO  sizeof(ALIGN_TYPE) * 16 // Size of alignment info
+
+class Minor;
+
+class Major : public LinkedList::Node {
+public:
+    void initialize(size_t pages)
```
micahswitzer: Why isn't this a constructor?

Kernel/Memory/Heap.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+    // Align the buffer size if necessary
+    if (pages % ARCH_PAGE_SIZE == 0) {
+        pages = pages / ARCH_PAGE_SIZE;
+    } else {
+        pages = pages / ARCH_PAGE_SIZE + 1;
+    }
+
+    // Ensure pages is at least equal to the minumum size;
+    if (pages < pageCount) {
```
micahswitzer, line 148: It's not clear that `pageCount` is a constant

Kernel/Memory/Heap.cpp
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    Major* major = (Major*)Memory::newPage(pages * ARCH_PAGE_SIZE - 1);
+    if (major == nullptr) {
+        // Out of memory.
+        // TODO: Should we panic or just return NULL?
+        panic("Out of memory!");
+    }
+
+    major->initialize(pages);
```
micahswitzer: Ah I see why you didn't use a constructor. I think we should be using "placement new" for this kind of thing.

https://stackoverflow.com/questions/222557/what-uses-are-there-for-placement-new

Kernel/Memory/Heap.cpp
Kfeavel: Because we have to request pages manually (no new or malloc by this point) so we have to manually cast the memory returned into a class.

Kernel/Memory/Heap.cpp
Kfeavel: Oh interesting. I'll look into that.

Kernel/Library/string.cpp
Kfeavel: Because that's how liballoc had it. You're right, probably should be size_t's though.

Kernel/Memory/Heap.cpp
Kfeavel: How would you denote that?

micahswitzer: Wow the style guide (https://webkit.org/code-style-guidelines/) is surprisingly unclear on that. I guess just mark it `const`.

Kernel/Library/string.cpp
micahswitzer: Yeah and I have no idea if they need to unaligned start/destination addresses, but that might be something to account for seeing as this is indented to be a generic `memcpy` replacement.

Kernel/Memory/paging.cpp
```diff
@@ -226,21 +226,30 @@ static uintptr_t findNextFreeVirtualAddress(size_t seq)
...
+            goto fail;
+        }
         Arch::Memory::Address phys(phys_page_idx * ARCH_PAGE_SIZE);
         Arch::Memory::Address vaddr(i * ARCH_PAGE_SIZE);
         mapKernelPage(vaddr, phys);
     }
+
+    page = (void*)(free_idx * ARCH_PAGE_SIZE);
+
+fail:
```
micahswitzer: This would be a good place to use RAII semantics to release the lock on return.

Kfeavel: Yeah, I was thinking that too. I was probably going to make a lambda function for mutexes and semaphores like we do for interrupt critical regions.

Kfeavel: Actually, what I'll probably do is write a wrapper class around Mutex that uses RAII. I could then write another wrapper class for lambda functionality, that way we have either one we want.

Kfeavel: Resolved by #370

Kernel/Applications/spinner.cpp
```diff
@@ -10,19 +10,37 @@
...
 void spinner(void) {
     Console::printf("\n");
     int i = 0;
     const char spinnay[] = { '|', '/', '-', '\\' };
+    for (size_t size = 1; size < 1024 * 1024 * 1024; size++) {
+        void* ptr = malloc(size);
+        if (ptr == nullptr) {
+            panic("null malloc in stress test!");
+        }
+        memset(ptr, 0xCAFEBABE, size / sizeof(0xCAFEBABE));
```
micahswitzer: This is only writing one quarter of the memory allocated since `memset` only writes bytes of data

Kernel/Applications/spinner.cpp
```diff
@@ -10,19 +10,37 @@
...
+#include <Arch/Memory.hpp>
+#include <Panic.hpp>
 
 namespace Apps {
 
 void spinner(void) {
     Console::printf("\n");
     int i = 0;
     const char spinnay[] = { '|', '/', '-', '\\' };
+    for (size_t size = 1; size < 1024 * 1024 * 1024; size++) {
```
micahswitzer: I'm offended that you didn't just write a new task for this test. Poor `spinner` doesn't deserve to be abused like this.

micahswitzer: It might also be nice to test this code outside of the kernel so that it's easier to debug.

Kfeavel: Reset the spinner in 8230d4fec34dadabde6c8370b30c4787170b2129. Its suffering is over.

Kfeavel:
Will write some unit tests for this soon. Thanks to @micahswitzer for writing a basic one used for testing ([`janky-heap-tests`](https://github.com/XyrisOS/xyris/tree/janky-heap-tests)).

PR #370, opened 2022-01-27
Resolve "RAII and Lambda Mutex Locking"

### LINKED ISSUE #369: RAII and Lambda Mutex Locking

Kfeavel:
Closes #369


Kernel/Locking/RAII.cpp
```diff
@@ -0,0 +1,22 @@
...
+ * @brief
+ * @version 0.1
+ * @date 2022-01-26
+ *
+ * @copyright Copyright the Xyris Contributors (c) 2022
+ *
+ */
+#include "RAII.hpp"
+
+RAIIMutex::RAIIMutex(Mutex& mutex)
```
micahswitzer: I really feel like this code should be in the header file. Otherwise it's going to add extra overhead.

PR #372, opened 2022-02-08
Resolve "Add Kernel Logger"

### LINKED ISSUE #357: Add Kernel Logger

Kfeavel:
Closes #357

### DIFF
--- a/Kernel/Bootloader/Handoff.cpp
+++ b/Kernel/Bootloader/Handoff.cpp
@@ -15,6 +15,7 @@
 #include <Library/stdio.hpp>
 #include <Library/string.hpp>
 #include <Memory/paging.hpp>
+#include <Logger.hpp>
 #include <Panic.hpp>
 // Generic devices
 #include <Devices/Graphics/console.hpp>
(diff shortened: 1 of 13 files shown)


Kfeavel:
As of right now the `Logger` constructor is not getting called at all and I'm at a complete loss as to why that would be. I've tried moving it into the global namespace, calling the constructor manually, printing inside it, and just about anything else. No indication that it's getting initialized. When I moved it to the global namespace, the constructor appeared in the `.ctors` section according to Ghidra, but `m_logLevel` was still `0` (`lTrace`) despite the code showing it should be set to `3` (`lInfo`).

**Edit:** I wonder if this is also why the prime number "program" is no longer working... If the bitset it uses isn't being defaulted to `1` instead of `0` then it would explain the regression. This could also be related to what order objects are being linked into the kernel. According to the [OSDev Wiki](https://wiki.osdev.org/Calling_Global_Constructors#Using_crti.o.2C_crtbegin.o.2C_crtend.o.2C_and_crtn.o_in_a_Kernel), linking `crti.o`, `crtbegin.o`, `crtn.o` and `crtend.o` must be done in a specific order otherwise odd behavior might occur. @micahswitzer is there a way we could force SCons to link those objects in a certain order?

Kfeavel:
Global constructor issue was resolved (at least in part) in #373. However, even though things like the primes task are now working properly, the static `Logger` instance is still not being constructed properly. The log level is being ignored.

Kfeavel:
@micahswitzer raises the point that having something to replace all of the `__func__` arguments would be nice. I propose that once we create the standard library directory, we can add a `SourceLocation` class that mimics the functionality of [`std::source_location`](https://en.cppreference.com/w/cpp/utility/source_location)

micahswitzer (approved):

PR #382, opened 2022-02-13
Resolve "Create XS Standard Library"

### LINKED ISSUE #378: Create `XS` Standard Library
Standard library source code (such as bitmaps, linked lists, etc.) should be moved to an `XS` directory in the root level of the project. It will be compiled as a static library that can be linked into the kernel or into the unit tests.

The overall design should be something similar to Serenity OS's `AK` standard library. A `__kernel__` and `__xyris__` define will need to be added to the build process to help differentiate between code usable within the kernel, userspace, and 
on other systems.

Additional Classes / Features to be added:

- [ ] `SourceLocation` ([`std::source_location`](https://en.cppreference.com/w/cpp/utility/source_location))
- [ ] `Vector` ([`std::vector`](https://en.cppreference.com/w/cpp/container/vector)) (#295)
- [ ] `String` ([`std::string`](https://en.cppreference.com/w/cpp/string/basic_string))

Kfeavel:
Resolves #378


Kernel/SConscript
```diff
@@ -6,6 +6,10 @@ env = env.Clone()
 linkerscript = env.File('Arch/$ARCH/linker.ld')
 env.Append(
     LINKSCRIPT=linkerscript,
+    CPPDEFINES={
+        '__xyris__': None,
+        '_kernel__': None,
```
jamesaorson, line 11: Missing a leading underscore

jamesaorson (changes requested):

PR #392, opened 2022-12-29
Update to the last 3.x release of Limine

micahswitzer:
We still need to update to the Limine boot protocol in order to use the latest 4.x release train


Kfeavel (approved):
lgtm :+1:

PR #397, opened 2023-01-02
Build limine-deploy in its own SConscript

micahswitzer:
This keeps the build artifacts out of the source tree


Kfeavel (approved):
:tada:

